PR #1, opened 2021-06-05
Check for null in LootingLevelEvent

vemerion:
Unfortunately, due to interactions with global loot modifiers, the damage source can be null in LootingLevelEvent. This fix avoids a potential null pointer exception by checking to make sure that the damage source is indeed not null.


leafreynolds:
Hey good catch! Appreciate your help! 😊

PR #10, opened 2022-10-25
Bondsmith Honorblade

Hawktrav:
The Bondsmith Honorblade now has a model and texture


leafreynolds:
I'll remove the shardplate changes after merging :)

PR #28, opened 2023-10-19
[Allomancy] Groundwork for iron and steel wall penetration

Gerbagel:

* Threaded block scanning for performance reasons (would have a minor stutter every 15 ticks otherwise, small but noticeable)
* Block penetration resistance calculations
* Less line clutter (which also leads to better performance, more than is lost by the previous point) 
* Pushing and pulling through blocks with resistances accounted for

### For the future
* Consider resistance in push/pull strength
* Consider resistance in drawing of lines


leafreynolds:
This is really cool 👀 will take a look through it all soon!

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
```diff
@@ -541,4 +570,221 @@ public static void CheckRecipeForMetal(TagKey<Item> containsMetal, Recipe<?> rec
...
+		@Override
+		@OnlyIn(Dist.CLIENT)
+		public void run() {
+			while (!isStopping) {
+				final Minecraft mc = Minecraft.getInstance();
+				final ProfilerFiller profiler = mc.getProfiler();
+				ScanResult nextScan;
+				LocalPlayer playerEntity = mc.player;
+				//only update box list every so often
+				// will run even if iron and steel are off. is problem? maybe
```
leafreynolds: Ideally we wouldn't run this code if the user isn't using steel. Maybe we can start/stop this thread as needed ?

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
leafreynolds, line 1: I don't think you meant to commit changes in this file :D

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
```diff
@@ -83,6 +116,10 @@ public void onModeChange(ISpiritweb cap, int lastMode)
 
 		if (getMode(cap) != 0)
 		{
+			if (getMode(cap) > 0)
+				startLinesThread();
+			else if (getMode(cap) < 0)
```
leafreynolds: We can use <= here, since 0 also means not burning

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
```diff
@@ -541,4 +574,220 @@ public static void CheckRecipeForMetal(TagKey<Item> containsMetal, Recipe<?> rec
...
+		@Override
+		@OnlyIn(Dist.CLIENT)
+		public void run() {
+			while (!isStopping) {
+				final Minecraft mc = Minecraft.getInstance();
+				final ProfilerFiller profiler = mc.getProfiler();
+				ScanResult nextScan;
+				LocalPlayer playerEntity = mc.player;
+				//only update box list every so often
+				if (playerEntity.tickCount % 15 != 0)
```
leafreynolds: I'm less familiar with threads, but would doing something like this be helpful?
`(playerEntity.tickCount + 5) % 15`, where 5 is whatever we feel like.

It's something I've been thinking adding elsewhere, which would help offload processing for different things onto different ticks.

If that doesn't matter, since this is already happening on a different thread, then feel free to ignore 😄

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
```diff
@@ -541,4 +574,220 @@ public static void CheckRecipeForMetal(TagKey<Item> containsMetal, Recipe<?> rec
...
+								final boolean validMetalBlock = block instanceof IHasMetalType iHasMetalType && iHasMetalType.getMetalType() != Metals.MetalType.ALUMINUM;
+								boolean isGood = validMetalBlock || containsMetal(block);
+
+								if (isGood)
+								{
+									try
+									{
+										Player player = Minecraft.getInstance().player;
+										Level level = Minecraft.getInstance().level;
+										BlockState endBlock = Objects.requireNonNull(level.getBlockState(blockPos));
```
leafreynolds, line 671: Somehow managed to crash my game when opening up the debug menu with graph via shift + f3.
image

Running the game again didn't crash it the second time, but I suspect it might be a race condition, as I still got an exception in thread. This might be happening because we push/pop stuff on the profiler from a different thread. We may need to remove those from the thread :D 

image

leafreynolds (changes requested):
I've been able to run these changes and it's looking pretty good! Hopefully removing the profiler push/pop from the thread will stop any weirdness from happening

To test, you may need to merge 1.19.x from upstream into your branch, as I fixed a jade crash with tooltip localisation missing.

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
Gerbagel: I did :)
Depending on the commit you're talking about, I either added or removed the thread start function there. Had it start on the entity joined event if the entity was a player. I pushed a commit last night that made it start when steel or allomancy is (mode > 0) instead

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
Gerbagel: I'm honestly not sure. Since it's a different thread, we might be able to remove the restriction entirely, but also, that might be too performance heavy? I decided to leave it up to you. I'll try the `playerEntity.tickCount + 5` and see if it makes a difference

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
Gerbagel: I'll put it in the start/stop methods instead then. Or should I remove it entirely? I'm not actually sure what it does, as I'm pretty new to Minecraft modding as a whole

Gerbagel: I know what it does now. I decided to merge the two profiler things into one, just called "cosmere-linesThread". I push it at the start of the while loop, and pop it at the end. After some professional debugging (spamming shift + f3 for a few seconds), it doesn't seem to crash anymore. I also can't see the thread on the pie chart, which is either really good, or it just isn't working.

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
Gerbagel: Tried playing with it. Either I have a super computer, or removing the restriction did almost nothing to performance. It was jarring to see highlights popping in and out though, so some restriction is probably necessary. I set it to `(playerEntity.tickCount + 5) % 5` for now, still a little jarring, but much better. I'd love to hear your thoughts once I've pushed.

Gerbagel:
### Commits summary

- Fixed thread/start stop issue, meaning I fixed the modes so the thread stops when the mode is <= 0. There were also more issues with how the thread started and stopped, fixed those too.
- Reduced lines thread frequency and also added the suggested offset. I see no difference in-game, but it might work slightly better. You'll have to judge for yourself.
- Fixed the profiler issue. It doesn't crash anymore when pressing shift + f3, but I also can't see profiler info in there, so I'm not sure it's working, but I hope it is. Negligible performance impact would be cool.

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
leafreynolds: ah good to know! I was just looking through the files that have changes in this pull request and it's basically just empty space and unused imports. If we can, would be good to remove those completely so that the file doesn't report changes

https://github.com/leafreynolds/cosmere/pull/28/files#diff-f62356cfe785e626b56b9fe94f844d391a0b4be1ed7c893439e390c434f43508

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
leafreynolds: aaaaaaaaaand ignore me, it's contained in an if statement that prevents 0

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
Gerbagel: Yeah, that makes sense. I totally forgot IntelliJ imports all that automatically and doesn't remove it. I'll scout through the files I've changed and remove any unused imports

Gerbagel: All done! I think!

leafreynolds: perfect thank ye~

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
leafreynolds: Maybe we should just remove it outright and see if anyone complains 👀 I have some peeps on the discord server with potato computers that we can get to test

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
leafreynolds: Looking good! I'll have a play around later today and see if I can get the profiler graph stuff to show. If it's just not usable in a thread, we can remove it.

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
Gerbagel: Would be interesting to see, at least. Worst case we can just hop on back to once per 15 ticks, just like before

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
leafreynolds: All right, can't find it in any of the menus. maybe it's incorrectly listing it as unspecified, but more likely it's just ignoring it since it's not from the same thread. It probably doesn't make sense to track it, since the profiler is specific to the thread it's tracking. 

Let's just get rid of the profiler stuff in the thread :D

Gerbagel: Alright! Gimme a sec and I'll have another commit pushed :)

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
leafreynolds: I'll get you to make that change, and then will make some builds off the branch and send a call out to the discord peeps 😎

Gerbagel: Done! Ready for the tater testers! I wish their PCs luck o7

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyIronSteel.java
Gerbagel: Also done!

leafreynolds (approved):
😎

PR #31, opened 2023-11-03
Add metalmind summary to powers menu

Gerbagel:
- A thread that scans the client's inventory and summarizes how much of each feruchemical metal is stored
- Same but for curios inventory
- Display summary in the powers menu


src/feruchemy/java/leaf/cosmere/feruchemy/client/utils/FeruchemyChargeThread.java
```diff
@@ -0,0 +1,153 @@
...
+                try {
+                    Thread.sleep(50);       // 20 ticks per 1000ms, 1000/20 = 50, rest for 1 tick
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+                continue;
+            }
+
+            // hashmap to keep track of each metal's f-charge in the inventory
+            final HashMap<Metals.MetalType, Double> metalmindCharges = new HashMap<>();
```
leafreynolds: We should probably save some garbage collection by reusing memory where possible, especially when we know we'll be needing it again every other tick

leafreynolds (changes requested):
Currently text goes off the screen if it's trying to write too many lines. We need to either start a new column, or restrict text on the left to be specific to the category they relate to. Allomancy to allomancy, feruchemy to feruchemy etc.

image

src/feruchemy/java/leaf/cosmere/feruchemy/client/utils/FeruchemyChargeThread.java
Gerbagel: Moved to the start of the thread instead, as we only need it while the thread is running

leafreynolds (changes requested):
Bendalloy and Cadmium appear to be missing squares in full manifestation type view. The user currently only has 6 allomantic powers, so I'm assuming cadmium and bendalloy are just playing up.

image

Feruchemy also shows no number when a mouse is not over a power,
image

But when a power is selected, it shows zero. We should probably be consistent in this case.
image


In single wheel view, we should probably not show the metal text unless mouse over on a specific power.

image

leafreynolds (approved):
👍

Issue #42, opened 2023-11-19
[Allomancy] A-Zinc Crash?


MoonlitJolteon:
Happens in Dev Environment

log: https://mclo.gs/p5WWKSt

leafreynolds:
Closing as part of #51

PR #51, opened 2023-11-21
More bugfixes 

Gerbagel:
- Fix for #42 (hopefully)
- Possible fix for metalmind summary thread sometimes not counting anymore.
- Switched back to old tin sound indicator texture (note icon)


src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBendalloy.java
```diff
@@ -27,21 +30,59 @@ public AllomancyBendalloy(Metals.MetalType metalType)
...
 		}
 
+		int mode = getMode(data);
 		String uuid = data.getLiving().getStringUUID();
-		if (mode > 0 && !playerThreadMap.containsKey(uuid))
+
+		// data thread management
 		{
-			playerThreadMap.put(uuid, new BendalloyThread(data));
+			if (mode > 0 && !playerThreadMap.containsKey(uuid))
```
leafreynolds: Mode will always be more than zero if we reach here, as this function is only called if the user is confirmed and paid for burning.

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBendalloy.java
```diff
@@ -27,21 +30,59 @@ public AllomancyBendalloy(Metals.MetalType metalType)
...
+			{
+				playerThreadMap.put(uuid, new BendalloyThread(data));
+			}
+
+			playerThreadMap.entrySet().removeIf(entry -> !entry.getValue().isRunning || AllomancyEntityThread.serverShutdown || entry.getValue() == null);
```
leafreynolds: if the user is not burning, we wont be able to reach here because this function is not called.

See AllomancyManifestation#tick

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBendalloy.java
```diff
@@ -27,21 +30,59 @@ public AllomancyBendalloy(Metals.MetalType metalType)
...
 		}
 
-		playerThreadMap.entrySet().removeIf(entry -> !entry.getValue().isRunning || AllomancyEntityThread.serverShutdown);
+		// data processing
+		{
+			// check if bendalloy is off or compounding
+			if (mode <= 0)
```
leafreynolds: same deal here, no need to check for off or compounding as user is confirmed normal burn mode at this point

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBendalloy.java
```diff
@@ -27,21 +30,59 @@ public AllomancyBendalloy(Metals.MetalType metalType)
...
+				return;
+			}
+
+			// this is the only way to check if the player is still online, thanks forge devs
+			if (data.getLiving().level.getServer().getPlayerList().getPlayer(data.getLiving().getUUID()) == null)
+			{
+				return;
+			}
+			//offload to power's tick
+			boolean isActiveTick = getActiveTick(data) % 20 == 0;
```
leafreynolds: it should always be active tick if they are in here as this it the applyEffectTick function

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBendalloy.java
```diff
@@ -27,21 +30,59 @@ public AllomancyBendalloy(Metals.MetalType metalType)
...
+
+			//Slows Down Time for the entities around the user
+			if (isActiveTick)
+			{
+				if (playerThreadMap.get(uuid) == null)
+				{
+					playerThreadMap.put(uuid, new BendalloyThread(data));
+				}
+				List<LivingEntity> entitiesToAffect = playerThreadMap.get(uuid).requestEntityList();
```
leafreynolds: perfect

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBrass.java
```diff
@@ -27,14 +30,67 @@ public AllomancyBrass(Metals.MetalType metalType)
...
 	{
 		int mode = getMode(data);
-
 		String uuid = data.getLiving().getStringUUID();
-		if (mode > 0 && !playerThreadMap.containsKey(uuid))
+
+		// data thread management
 		{
-			playerThreadMap.put(uuid, new BrassThread(data));
+			if (mode > 0 && !playerThreadMap.containsKey(uuid))
```
leafreynolds: Same as bendalloy comments, we should probably override tick, and stick this stuff in there. After that, we can have it call the parent class tick

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBrass.java
```diff
@@ -27,14 +30,67 @@ public AllomancyBrass(Metals.MetalType metalType)
...
+			}
+
+			playerThreadMap.entrySet().removeIf(entry -> !entry.getValue().isRunning || AllomancyEntityThread.serverShutdown);
 		}
 
-		playerThreadMap.entrySet().removeIf(entry -> !entry.getValue().isRunning || AllomancyEntityThread.serverShutdown);
+		// data processing
+		{
+			// check if brass is off or compounding
+			if (mode <= 0)
```
leafreynolds: Same as other comments, we already know mode is more than zero here

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBrass.java
```diff
@@ -51,66 +107,43 @@ public BrassThread(ISpiritweb data)
...
-
-					// this is the only way to check if the player is still online, thanks forge devs
-					if (data.getLiving().level.getServer().getPlayerList().getPlayer(data.getLiving().getUUID()) == null)
-					{
-						break;
-					}
 
-					//todo, replace x * mode with config based value
-					double allomanticStrength = getStrength(data, false);
+				if (mode <= 0)
```
leafreynolds, line 125: perfect

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyCadmium.java
```diff
@@ -29,14 +32,82 @@ public AllomancyCadmium(Metals.MetalType metalType)
...
+			}
+
+			playerThreadMap.entrySet().removeIf(entry -> !entry.getValue().isRunning || AllomancyEntityThread.serverShutdown || entry.getValue() == null);
 		}
 
-		playerThreadMap.entrySet().removeIf(entry -> !entry.getValue().isRunning || AllomancyEntityThread.serverShutdown);
+		// data processing
+		{
+			// check if cadmium is off or compounding
+			if (mode <= 0)
```
leafreynolds: Same as other comments, we already know mode is more than zero here

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyCadmium.java
```diff
@@ -29,14 +32,82 @@ public AllomancyCadmium(Metals.MetalType metalType)
 	protected void applyEffectTick(ISpiritweb data)
 	{
 		int mode = getMode(data);
-
 		String uuid = data.getLiving().getStringUUID();
-		if (mode > 0 && !playerThreadMap.containsKey(uuid))
+
+		// data thread management
 		{
```
leafreynolds: Same as bendalloy comments, we should probably override tick, and stick this stuff in there. After that, we can have it call the parent class tick

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyZinc.java
```diff
@@ -29,14 +32,74 @@ public AllomancyZinc(Metals.MetalType metalType)
 	protected void applyEffectTick(ISpiritweb data)
 	{
 		int mode = getMode(data);
-
 		String uuid = data.getLiving().getStringUUID();
-		if (mode > 0 && !playerThreadMap.containsKey(uuid))
+
+		// data thread management
 		{
```
leafreynolds: Same as bendalloy comments, we should probably override tick, and stick this stuff in there. After that, we can have it call the parent class tick

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyZinc.java
```diff
@@ -29,14 +32,74 @@ public AllomancyZinc(Metals.MetalType metalType)
...
+			}
+
+			playerThreadMap.entrySet().removeIf(entry -> !entry.getValue().isRunning || AllomancyEntityThread.serverShutdown || entry.getValue() == null);
 		}
 
-		playerThreadMap.entrySet().removeIf(entry -> !entry.getValue().isRunning || AllomancyEntityThread.serverShutdown);
+		// data processing
+		{
+			// check if zinc is off or compounding
+			if (mode <= 0)
```
leafreynolds: Same as other comments, we already know mode is more than zero here

src/feruchemy/java/leaf/cosmere/feruchemy/client/utils/FeruchemyChargeThread.java
leafreynolds, line 1: In the start function, what happens to the last thread if it is not null? is there a way to check if it's doing anything and stop it/kill the thread to be extra sure?

src/feruchemy/java/leaf/cosmere/feruchemy/client/utils/FeruchemyChargeThread.java
```diff
@@ -73,83 +73,95 @@ public void run() {
...
-                        return true;
-                    });
+                            return true;
+                        });
 
-            lock.lock();
-            try {
-                feruchemyChargeMap.clear();
-                feruchemyChargeMap.putAll(metalmindCharges);
+                lock.lock();
```
leafreynolds: We should change this from lock to if(lock.tryLock()), so that the thread doesn't throw an exception when the main thread is accessing the info. That way it just skips when the main thread is using it.

We may also want to just force the main thread to check on a certain tick number and the count thread to check on another. Being extra sure they aren't interfering with eachother and still able to do their own jobs.

src/feruchemy/java/leaf/cosmere/feruchemy/client/utils/FeruchemyChargeThread.java
```diff
@@ -73,83 +73,95 @@ public void run() {
...
-                    throw new RuntimeException(e);
+                if (mc.player.tickCount % 2 != 0)
+                {
+                    try
+                    {
+                        Thread.sleep(50);       // 20 ticks per 1000ms, 1000/20 = 50, rest for 1 tick
+                    }
+                    catch (InterruptedException e)
+                    {
+                        throw new RuntimeException(e);
```
leafreynolds: if an exception is thrown here, what happens? The object wont have set itself to isStopping, right?

leafreynolds (changes requested):

src/feruchemy/java/leaf/cosmere/feruchemy/client/utils/FeruchemyChargeThread.java
Gerbagel: Unfortunately, I don't think so. But it shouldn't need to. The thread will kill itself whenever the player leaves a world, and should always either be null or not running whenever the player next joins a world

src/feruchemy/java/leaf/cosmere/feruchemy/client/utils/FeruchemyChargeThread.java
Gerbagel: We could make the thread check every odd tick, and the main thread every even tick?

src/feruchemy/java/leaf/cosmere/feruchemy/client/utils/FeruchemyChargeThread.java
Gerbagel: I don't actually remember, and I don't know why that throw is there. It just seemed important since it's so different from the other threads that I didn't touch it. I think I'll just get rid of it though

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBendalloy.java
Gerbagel: Strangely, somehow, I think that's your code. I don't comment like that, so probably a remnant? Will remove either way :)

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBrass.java
```diff
@@ -27,14 +30,67 @@ public AllomancyBrass(Metals.MetalType metalType)
...
+				}
+
+				List<LivingEntity> entitiesToAffect = playerThreadMap.get(uuid).requestEntityList();
+				for (LivingEntity e : entitiesToAffect)
+				{
+					if (e instanceof Mob mob)
+					{
+						mob.setNoAi(mode == 3 && allomanticStrength > 15);
+
+						switch (mode)
```
Gerbagel: Just realised, this switch case might have to do with #64. There's no breaks, unless that's intended?

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyCadmium.java
Gerbagel: I assume you mean all the data processing stuff? I think that's probably a good idea, if so

src/feruchemy/java/leaf/cosmere/feruchemy/client/utils/FeruchemyChargeThread.java
leafreynolds: I had a thought earlier, I'm unsure how guaranteed it is for the tick to not change mid-operation, but I imagine it would help somewhat, especially when combined with try lock. 
More likely is the main thread ticking over and then trying to access and iterate the list while the thread is still building it

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBrass.java
leafreynolds: The original intention was to be able to have it do all of them depending on mode, flaring would still turn off aggressiveness and forget the last attacker.

This still needs changing, because it's obviously not working correctly.

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBendalloy.java
leafreynolds: yeah, sorry to make you clean up after me! 😅

src/allomancy/java/leaf/cosmere/allomancy/common/manifestation/AllomancyBrass.java
Gerbagel: Alright, I'll leave it for a future PR

src/feruchemy/java/leaf/cosmere/feruchemy/client/utils/FeruchemyChargeThread.java
Gerbagel: So that's a go on the odd-even tick split? Or just the tryLock()?

leafreynolds: Let's do both, so at least if the main thread gets to it first, the charge thread doesn't needlessly try modify.

Issue #78, opened 2023-12-07
[Feruchemy] Gold TLC


leafreynolds:
# How it currently works
How much you tap determines the number of points of health you get back, which happens once per second.

# Changes Proposed
Higher tap rate should mean _faster_ healing, not high amounts of health per one second. Health restored will depend on minimum and maximum possible tap rates. Minimum will be something like half a heart every 10-20 seconds. Maximum tap rate will be half a heart every _tick_. This should be _very_ expensive. It should be possible for the player tap rate to outstrip even max strength compounding.

Additionally, I would like to consider a potential nerf. Rather than stop tapping when at full health, lets give them a buff of temporary health similar to eating a golden apple when you get the gold hearts. This will only be available at high enough tap rates, lower rates will still cost stored attribute for no extra gain.

This will force the player to need to be more strategic about their tapping of gold. This may or may not be a good gameplay decision, but would nicely fit how Miles talks about how tapping gold all the time always makes him feel good.

Testing will be required.

PR #89, opened 2024-02-21
[Feruchemy] Gold TLC

Gerbagel:
- Gold now heals faster rather than more hearts, half a heart per "heal tick"
- Cost is higher
- Gold no longer stops tapping at max health, and instead grants extra hearts above a certain tap level
- Death resistance now has a cost, and leaves the player with 1 health remaining after lethal damage


Gerbagel:
Related issue: #78

PR #92, opened 2024-02-27
Metalborn Powers Choice

Gerbagel:

- Added commands to choose which metalborn powers you want on spawn
- Added a server-side config to turn this feature off or on (default off)


src/main/java/leaf/cosmere/common/eventHandlers/EntityEventHandler.java
```diff
@@ -67,9 +71,23 @@ public static void onEntityJoinWorldEvent(EntityJoinLevelEvent event)
...
+					spiritweb.setHasBeenInitialized();
+				}
+				else
+				{
+					// spiritweb not initialized yet in this case
+					Player player = (Player) eventEntity;
+
+					String command = "/cosmere choose_metalborn_powers ";
+					MutableComponent instructionComponent = Component.literal("To choose powers, use ");
+					instructionComponent.append(Component.literal("§6§n/cosmere choose_metalborn_powers [allomancy] [feruchemy]")
```
leafreynolds, line 87: Let's make this a command that they can actually click on to trigger for them

Gerbagel: It already is, isn't it? You click on it and it puts "/cosmere choose_metalborn powers " in the player's command space thing, and they can fill in the rest themselves

leafreynolds: d'oh.... I missed the line literally underneath it calling it a with click event

Never mind!

leafreynolds:
Can we add another command that lets an admin reset a user's powers and initialized state? Would let the user try again with minimal effort from an admin (or if we missed something in the code that prevents things working properly)

Gerbagel:
> Can we add another command that lets an admin reset a user's powers and initialized state? Would let the user try again with minimal effort from an admin (or if we missed something in the code that prevents things working properly)

I can do that, shouldn't be too hard

Gerbagel:
Added the command :)

Gerbagel:
Also putting the translation key fix here, mostly for convenience

PR #145, opened 2025-01-28
Activator Keybinds

SoaringEaqle:
Adding keybinds for allomancy, feruchemy, and surgebinding.

When using the keybind, the keybound manifestation becomes the active one. For allomancy, when "shift" is held allomancy will flare, with "control" it will compound, and will do both when both are held. For feruchemy, shift sets it to level 5.

If the power is already active, it will shut off.

Surgebinding doesn't do anything yet.


SoaringEaqle:
I may have also accidentally thrown in some patchoui stuff.

PR #149, opened 2025-02-10
[Surgebinding] SL Keybinds & SL Drawing/Breathing

Roseplate:
I think this works but also please look over if you can <3 I do not trust myself even a bit 😭 .

Added functions in spiritweb for drawing in and breathing out stormlight.
Both of those call on functions that draw from and breathe into specified gems.

The aforementioned are called by packets sent out by keybinds. :3
Breathe Stormlight / Breathe key refers to breathing out stormlight.
Draw Stormlight / Draw key refers to drawing in stormlight.

The only untested bit is the herald-case but it's pretty much as it was and drawing in stormlight in a highstorm.

### DIFF
--- a/src/surgebinding/java/leaf/cosmere/surgebinding/client/SurgebindingKeybindings.java
+++ b/src/surgebinding/java/leaf/cosmere/surgebinding/client/SurgebindingKeybindings.java
@@ -26,13 +26,13 @@ import java.util.stream.Collectors;
 import static leaf.cosmere.api.Constants.Strings.*;
 import static leaf.cosmere.surgebinding.common.registries.SurgebindingManifestations.SURGEBINDING_POWERS;
 
-
 // Really only has its own file to more nicely reference keybindings.
 // Otherwise, could have lived in mod client events
 @Mod.EventBusSubscriber(value = Dist.CLIENT, modid = Surgebinding.MODID, bus = Bus.MOD)
 public class SurgebindingKeybindings
 {
 	public static KeyMapping SHARDBLADE;
+	public static KeyMapping DRAW_STORMLIGHT;
 	public static KeyMapping BREATHE_STORMLIGHT;
 
 	public static final Map<Roshar.Surges,KeyMapping> SURGEBINDING_POWER =
@@ -60,7 +60,8 @@ public class SurgebindingKeybindings
 		}
 
 		event.register(SHARDBLADE = new KeyMapping(KEY_SHARDBLADE, GLFW.GLFW_KEY_X, "keys.surgebinding.main"));
-		event.register(BREATHE_STORMLIGHT = new KeyMapping(KEY_BREATHE_STORMLIGHT, GLFW.GLFW_KEY_Z, "keys.surgebinding.main"));
+        event.register(DRAW_STORMLIGHT = new KeyMapping(KEY_DRAW_STORMLIGHT, GLFW.GLFW_KEY_Z, "keys.surgebinding.main"));
+		event.register(BREATHE_STORMLIGHT = new KeyMapping(KEY_BREATHE_STORMLIGHT, GLFW.GLFW_KEY_Y, "keys.surgebinding.main"));
 
 	}
 
--- a/src/surgebinding/java/leaf/cosmere/surgebinding/client/SurgebindingKeybindings.java
+++ b/src/surgebinding/java/leaf/cosmere/surgebinding/client/SurgebindingKeybindings.java
@@ -32,8 +32,8 @@ import static leaf.cosmere.surgebinding.common.registries.SurgebindingManifestat
 public class SurgebindingKeybindings
 {
 	public static KeyMapping SHARDBLADE;
-	public static KeyMapping DRAW_STORMLIGHT;
-	public static KeyMapping BREATHE_STORMLIGHT;
+	public static KeyMapping REQUEST_STORMLIGHT;
+	public static KeyMapping DISPATCH_STORMLIGHT;
 
 	public static final Map<Roshar.Surges,KeyMapping> SURGEBINDING_POWER =
 			Arrays.stream(EnumUtils.SURGES)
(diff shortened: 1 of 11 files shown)


src/surgebinding/java/leaf/cosmere/surgebinding/client/SurgebindingKeybindings.java
```diff
@@ -60,7 +60,8 @@ public static void register(RegisterKeyMappingsEvent event)
 		}
 
 		event.register(SHARDBLADE = new KeyMapping(KEY_SHARDBLADE, GLFW.GLFW_KEY_X, "keys.surgebinding.main"));
-		event.register(BREATHE_STORMLIGHT = new KeyMapping(KEY_BREATHE_STORMLIGHT, GLFW.GLFW_KEY_Z, "keys.surgebinding.main"));
```
Roseplate, line 63: Breathe used to refer to drawing in stormlight but I use "Draw" for that so I changed this to Y and made it refer to breathing out and gave Z to draw.
I would've made it default to "Shift + Z" but I wasn't sure how to and google wasn't helping so ah well :P

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/SurgebindingSpiritwebSubmodule.java
```diff
@@ -346,13 +346,9 @@ public void breatheStormlight() {
...
 	public static boolean isInHighstorm(LivingEntity entity) {
-		if (entity.level().isThundering()
+		return entity.level().isThundering()
 				&& entity.level().dimension().equals(SurgebindingDimensions.ROSHAR_DIM_KEY)
-				&& entity.level().isRainingAt(entity.blockPosition()))
-		{
-			return true;
-		}
-		return false;
+				&& entity.level().isRainingAt(entity.blockPosition());
```
Gerbagel, line 365: Not to be picky, but isn't this redundant? If it's thundering, it should already be raining, right? Maybe I don't understand Minecraft weather, that's possible

Roseplate: This is taken from the previous if statement that was already in, I just moved it. But either way it's different from checking whether it's thundering. It's checking whether there is rain at a specific spot, since you shouldn't be able to draw in stormlight if you're indoors or underground. So basically it's "if outdoors && thundering". but this is only a temporary method until a proper highstorm gets added, I just feel bad deleting other people's work.

leafreynolds: Pretty much like Rose said. The intention was for players needing to be outside to get infused. Otherwise it would work even on players deep underground.

src/surgebinding/java/leaf/cosmere/surgebinding/client/SurgebindingKeybindings.java
leafreynolds: maybe we could change it to request/dispatch rather than draw and breathe. Since the books typically refer to it as 'breathing in stormlight', we don't want to introduce confusion unnecessarily

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/SurgebindingSpiritwebSubmodule.java
```diff
@@ -195,6 +203,182 @@ else if (livingEntity.isUnderWater())
 		}
 	}
 
+
+	private void drawGem(ItemStack item, int amountDrawn) {
```
leafreynolds: The project code styling typically puts `{` on new lines, which you've done in some places, but not others. 

Hopefully, if you merge from upstream, it should give you the styling enforce file thing I committed recently, then hopefully you can just go into each of those files and run the Code > Reformat Code (or similar) from the menu bar.

leafreynolds (changes requested):

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/SurgebindingSpiritwebSubmodule.java
Roseplate: I had no idea you could do that. Also, this is why you triple-read *everything* before starting to do stuff because I thought you meant when it was merged maybe it might get fixed so I just spent 10-20 minutes going through the code trying to catch what I can just in case it didn't (which it wouldn't since that wasn't what you we're saying at all), not to merge from upstream (it didn't occur to me because I already had xT) and use an IntelliJ feature I hadn't heard of 🤦‍♀️I am a moron sometimes XP anyway, it's done now <3

Roseplate: I think I caught most cases where I used Draw and Breathe and turned them into Request and Dispatch, at least in the important parts where it might get confusing. I left "drawSpeed" as it was since it's used for both.

leafreynolds: sorry to bamboozle yah, I was being lazy and figured you'd be okay going through them for clean up 😄 I'm happy to do a pass on it all later too.

leafreynolds:
thank you for your hard work :)

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/SurgebindingSpiritwebSubmodule.java
Roseplate: Don't worry I ended up learning something so I'm pretty grateful! ctrl + alt + shift + L to reformat code x3

PR #150, opened 2025-03-03
Nicrosil stopgap overhauls

B1ackN0va:
-added exception to f-nicrosil with leaf's help to allow the burning of a-nicrosil while storing
-changed a-nicrosil's min burn to -1, to allow compounding


B1ackN0va:
Metalmind and Nicrosil code changed to allow nicrosil to store powers as an nbt on the metalminds themselves, allowing stacking of different powers rather than overwriting
-Still a major bug with the nbt tag not clearing once the metalmind is drained

Issue #119, opened 2024-11-11
Keybinds for powers, activating and deactivating


leafreynolds:
We should add a keybind for each power (no default set) so that people can tap it to activate, again to deactivate if active. 

For allomancy, there should be two, one for flaring and one for normal burn.

Feruchemy undecided on default tap level. Could just go for max?

SoaringEaqle:
Just a few thoughts: 
First, If you're also holding shift (for example) when you press an Allomancy activation keybind would set it to compound, holding alt (for example) would set it to flare, holding both while pressing the activate would flare-compound. 
Second, you could reuse those key-binds for Feruchemy. You could have two tap/store levels configs (defaults of 4 and 15), then when you press the activate bind, it turns on the Feruchemic power to the first default, and if you press it while holding alt, it your power to tap at the second default or max (whichever is lower). In addition, if your hold shift, it will store at the respective levels.
Third, pressing the keybind while your power is already active (burning, flaring, compounding, tapping or storing), it should turn it off.

SoaringEaqle:
Actually, could I work on this?

leafreynolds:
> Actually, could I work on this?

Any time you see a task that isn’t actively being worked on, feel free to go ahead. Make a pull request too, linking to the ticket

leafreynolds:
> Actually, could I work on this?

Did you get a chance to work on this? :)

SoaringEaqle:
I started on it, so I believe all the constants and keybindings for Allomancy are in, I just needed to begin implementation. I got busy and kinda forgot, so thanks for reminding me I was doing this.

SoaringEaqle:
> Just a few thoughts: First, If you're also holding shift (for example) when you press an Allomancy activation keybind would set it to compound, holding alt (for example) would set it to flare, holding both while pressing the activate would flare-compound. Second, you could reuse those key-binds for Feruchemy. You could have two tap/store levels configs (defaults of 4 and 15), then when you press the activate bind, it turns on the Feruchemic power to the first default, and if you press it while holding alt, it your power to tap at the second default or max (whichever is lower). In addition, if your hold shift, it will store at the respective levels. Third, pressing the keybind while your power is already active (burning, flaring, compounding, tapping or storing), it should turn it off.

Before I start, does this work for you? Any changes you would like made?

leafreynolds:
> Before I start, does this work for you? Any changes you would like made?

[...]

Issue #106, opened 2024-05-30
Power-Mode save states - Keybind compat


leafreynolds:
There's tons of powers in the mod suite and it can be very annoying to need to turn multiple on/off. 

As a player, it would be convenient to push one keybind and have it save all the modes each power is set to. Then later, be able to push a different keybind and have it set all powers to those modes.

Additionally, it would be nice to roll this into the UI updates we would also like to do. This could mean selectively setting only a few powers to specific modes, rather than every power.

SoaringEaqle:
I had an idea for this utilizing a command. A player could use the following commands to add or remove powers to a mass activator. "/cosmere activator [#] add [power type] [level]" or "/cosmere activator [#] remove [power type]," where [#] is the mass activator number (1-4), and the other two are pretty self-explanatory. 

I can get started on making the activators even if you don't want the command-based implementation.

PR #152, opened 2025-05-03
Saves power states and removes keybind modifiers.

SoaringEaqle:

Removes unappreciated modifier keybinds for individual activators

Adds power save states.

Power save states can be saved by holding the "save power save" key and pressing a hotbar slot key, while having at least one manifestation active.

Power save states can be activated by holding the "activate power save" key and pressing the hotbar key tied to the power save you wish to recall.

## Future Plans for These Features:
Chat message when power save state is saved or activated 
Save state toggle, turning those powers back off after use.


leafreynolds:
Related discussion on discord
https://discord.com/channels/828383636465123328/1351674880831262860

SoaringEaqle:
Should resolve the following issues:
#119 
#106

Gerbagel:
I really should have actually checked this ages ago, I'm *very* sorry.

Anyway, found a bug. 

1. Save a state
2. Load that state
3. Save a different state to the same slot
4. Load again, the state will unload instead
5. Attempt to load again, only some changes will be reflected. When testing, I set f-gold to store at 1, and a-steel to flare, and only a-steel changed while f-gold remained at 0

Also, I'm almost certain this is just me struggling with datagen again, but I'm only getting translation strings for each of the individual power hotkeys as well as the activate and save hotkeys. Could you confirm it works on your end?

SoaringEaqle:
I never did translations. I’ll tackle that with the bug fixes. I’ll also add a patchouli guide for navigating the keybinds.

SoaringEaqle:
Done with the bug fixes.

Gerbagel:
Works mostly as expected, though it still has to unload before loading a save state again. 

Looking at the code, you either need to set `isActive` to false when the state changes, or check if the current state is different from the state the player is attempting to load.

SoaringEaqle:
I think I'm finally done. It might break on multiplayer, but I can't really test that.

leafreynolds:
> I think I'm finally done. It might break on multiplayer, but I can't really test that.

doesn't it let you runServer and runClient to connect to it?

SoaringEaqle:
I don't think I can test multiple players. The only reason I say it might break is because the power save states are copied to the sprirtweb as nbt data, and I have no clue how the server and client interact with that use case. It probably works, I just don't know how.

SoaringEaqle:
Upon further reflection and learning, I can say with 97% confidence that this will have zero issues on multiplayer.

Gerbagel:
No more issues that I can see. Multiplayer didn't crash, and it seems synced. Should be fine to merge if Leaf has nothing else.

PR #159, opened 2025-07-24
Mistcloak v2

SoaringEaqle:
Fixes mistcloak render by making it a Curios item to go in the head slot.


SoaringEaqle:
There's some stuff that still needs done, but it's mostly to do with how radiant shardplate works and interacts with radiants, so it's probably best to add it after radiants get added. Instead of working around a system that barely exists yet.

SoaringEaqle:
Needed a break from Investiture System, so mostly redid this. Hopefully this redesign makes it easier to influence data, and manipulate deadeye states, as well as control bonding easier.

Gerbagel:
Don't see any issues with the Mistcloak. 

However, when I go to change out one piece of Shardplate for another, the second one becomes the same type as the first. This happens regardless of Dead or living plate. 

`SurgebindingItemModelsGen.java` is also missing a couple imports for `GemstoneItem` and `ShardplateCurioItem`. 

I also get `java.lang.IllegalArgumentException: Texture surgebinding:item/gemstone_chip does not exist in any known resource pack` when I try to `runData`, which I was doing to see if the lack of a Curios slot name for Shardplate was just because I hadn't generated.

SoaringEaqle:
I did notice that, however, I do not know how to fix it. My assumption is that it is a quirk with curios. I think the best way to bypass it would be to store the data somewhere when you try to equip, and come back a few ticks later and reattach the data, but I really have no clue what that would look like.

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/HonorbladeItem.java
```diff
@@ -63,10 +75,49 @@ public Multimap<Attribute, AttributeModifier> getAttributeModifiers(EquipmentSlo
...
+		IShard data = stack.getCapability(ShardData.SHARD_DATA).orElseGet(() -> {return new ShardData(stack);});
+		data.setOrder(radiantOrder);
+		return stack;
+	}
+
+	@Override
+	public void inventoryTick(ItemStack pStack, Level pLevel, Entity pEntity, int pItemSlot, boolean pIsSelected)
+	{
+		super.inventoryTick(pStack, pLevel, pEntity, pItemSlot, pIsSelected);
+		getShardData(pStack).setOrder(getOrder(pStack));
```
leafreynolds: ideally we don't want this getting called every single tick

leafreynolds:
In future, we also don't want multiple features packed into a single pull request. This one is called mistcloak, but is mostly full of surgebinding stuff.

SoaringEaqle:
Sorry about that. I just needed a bunch of stuff I made for the mistcloak for the Shardplate, and I didn't want to go through the effort of duplicating the branch and removing the mistcloak-specific code.

SoaringEaqle:
> `SurgebindingItemModelsGen.java` is also missing a couple imports for `GemstoneItem` and `ShardplateCurioItem`.
> 
> I also get `java.lang.IllegalArgumentException: Texture surgebinding:item/gemstone_chip does not exist in any known resource pack` when I try to `runData`, which I was doing to see if the lack of a Curios slot name for Shardplate was just because I hadn't generated.

These two issues should be resolved. It was some issue where the files half updated, but not completely.

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/HonorbladeItem.java
leafreynolds: ideally we don't want this getting called every single tick

src/surgebinding/java/leaf/cosmere/surgebinding/common/entity/spren/Honorspren.java
```diff
@@ -35,7 +35,7 @@ public boolean canBeHitByProjectile()
 	@Override
 	public boolean canBeSeenByAnyone()
 	{
-		//TODO: make this only visible to the player it is bonded to?
+		//TODO: make this only visible to the player it is bond to?
```
leafreynolds: "is bond to" isn't good english 😅 also please try not to commit changes like these where you're only changing a comment

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/ShardbladeItem.java
```diff
@@ -54,9 +74,100 @@ public boolean isFoil(ItemStack itemStack)
...
+		ShardData data = getShardData(pStack);
+		if(pEntity instanceof Player player)
+		{
+			if(data.isLiving())
+			{
+				bond(pStack, player);
+			}
+			else if(data.bondTicks() >= bondTime())
+			{
+				bond(pStack, player);
```
leafreynolds: I'm not sure I understand this mechanic right for shardplate. You don't really bond plate like you would a radiant spren. You kinda have the plate spren, or you don't

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/ShardplateCurioItem.java
```diff
@@ -0,0 +1,396 @@
...
+	}
+
+	@Override
+	public void onEquip(SlotContext slotContext, ItemStack prevStack, ItemStack stack)
+	{
+		if (slotContext.entity().level().isClientSide) return;
+
+		// Copy the capability data if present
+		prevStack.getCapability(ShardData.SHARD_DATA).ifPresent(fromCap -> {
+			stack.getCapability(ShardData.SHARD_DATA).ifPresent(toCap -> {
```
leafreynolds: Why are we copying the capability from old stack to new stack? does it get lost somehow?

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/ShardplateCurioItem.java
```diff
@@ -0,0 +1,396 @@
...
+	@Override
+	public void onUnequip(SlotContext slotContext, ItemStack newStack, ItemStack stack)
+	{
+		if (slotContext.entity().level().isClientSide) return;
+
+		// Copy the capability data if present
+		stack.getCapability(ShardData.SHARD_DATA).ifPresent(fromCap -> {
+			newStack.getCapability(ShardData.SHARD_DATA).ifPresent(toCap -> {
+				CompoundTag nbt = fromCap.serializeNBT();
+				toCap.deserializeNBT(nbt);
```
leafreynolds: same here, a bit confused.

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/ShardplateCurioItem.java
```diff
@@ -0,0 +1,396 @@
...
+		}
+		else
+		{
+			return Roshar.getDeadplate();
+		}
+	}
+
+
+	@Override
+	public Multimap<Attribute, AttributeModifier> getAttributeModifiers(SlotContext slotContext, UUID uuid, ItemStack stack)
```
leafreynolds: do the attribute modifiers reset properly when the plate runs out of charge? or does it stick around till you equip and unequip it

leafreynolds (changes requested):
Apparently I never pushed submit on my review 🙃

leafreynolds:
For good measure, I wanna see what happens if I request additional review from @claude . Maybe it will catch something I haven't.

leafreynolds:
Nothing happens 🙄 I'll figure it out later. Claude's throwing some error for whatever reason.

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/ShardplateCurioItem.java
Gerbagel: Perhaps this has something to do with my previous comment? 

https://github.com/leafreynolds/cosmere/pull/159#issuecomment-3535051024

leafreynolds: Oh probably. These comments on my review were from a while ago, I just didn’t push submit apparently. 

I’ll take a proper look tomorrow.

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
```diff
@@ -110,5 +124,54 @@ public static void onLivingHurtEvent(LivingHurtEvent event)
...
+						}
+
+						float original = event.getAmount();
+						float absorbed = original * 0.2f;
+						float remaining = original - absorbed;
+
+						// Reduce damage taken by entity
+						event.setAmount(remaining);
+
+						stack.setDamageValue((int) (stack.getDamageValue() - absorbed));
```
leafreynolds: Pretty sure damage value isn't 'item health', it's how much damage it's taken. 0 = full/max=broken. so subtracting would make it 'heal'.

You  might need to test this one in game, break point it maybe and see what it's doing.

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
```diff
@@ -110,5 +124,54 @@ public static void onLivingHurtEvent(LivingHurtEvent event)
...
+		if (entity.level().isClientSide || event.getAmount() <= 0)
+		{
+			return;
+		}
+		// Only do damage types that the mistcloak can protect.
+		for(ResourceKey<DamageType> type: protectedDamageTypes)
+		{
+			if(source.is(type))
+			{
+				Optional<ICuriosItemHandler> sub = CuriosApi.getCuriosInventory(entity).resolve();
```
leafreynolds: Is this var used for anything ? otherwise it's a slightly more expensive call

src/api/java/leaf/cosmere/api/helpers/CuriosHelper.java
```diff
@@ -0,0 +1,132 @@
...
+public class CuriosHelper
+{
+	public static ICuriosItemHandler getCuriosHandler(LivingEntity entity)
+	{
+		return CuriosApi.getCuriosInventory(entity).resolve().isPresent()? CuriosApi.getCuriosInventory(entity).resolve().get(): null;
+	}
+
+	public static SlotResult getSlot(LivingEntity entity, String slotID, int index) {
+		ICuriosItemHandler sub = getCuriosHandler(entity);
+		assert sub != null;
```
leafreynolds: Assertions are disabled at runtime by default in Java, so should probs replace this with a null check if that's what you were wanting

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
```diff
@@ -110,5 +124,54 @@ public static void onLivingHurtEvent(LivingHurtEvent event)
...
+				if (CuriosHelper.getCuriosHandler(entity) != null)
+				{
+
+					for (SlotResult slotResult : CuriosHelper.getSlotsWithItem(entity, AllomancyItems.MISTCLOAK.asItem()))
+					{
+						ItemStack stack = slotResult.stack();
+						MistcloakItem item = (MistcloakItem) stack.getItem();
+						if (item.getDamage(stack) == 0)
```
leafreynolds: as with other comment, 0 would be max durability I think? needs checking

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
```diff
@@ -110,5 +124,54 @@ public static void onLivingHurtEvent(LivingHurtEvent event)
...
+		// Only do damage types that the mistcloak can protect.
+		for(ResourceKey<DamageType> type: protectedDamageTypes)
+		{
+			if(source.is(type))
+			{
+				Optional<ICuriosItemHandler> sub = CuriosApi.getCuriosInventory(entity).resolve();
+				if (CuriosHelper.getCuriosHandler(entity) != null)
+				{
+
+					for (SlotResult slotResult : CuriosHelper.getSlotsWithItem(entity, AllomancyItems.MISTCLOAK.asItem()))
```
leafreynolds: we should probably guard against multiple mistcloaks being equippable somehow. Or maybe break out of the loop after reducing the first bit of damage, so people can't double up.

.DS_Store
leafreynolds, line 1: Can you delete all these ds store files? If you merge in develop, the should stop trying to add themselves in future by accident.

src/allomancy/java/leaf/cosmere/allomancy/common/items/MistcloakItem.java
```diff
@@ -29,7 +40,27 @@ public final String getArmorTexture(ItemStack stack, Entity entity, EquipmentSlo
...
+	}
+
+	public Multimap<Attribute, AttributeModifier> getAttributeModifiers(SlotContext slotContext, UUID uuid, ItemStack stack)
+	{
+		ImmutableMultimap.Builder<Attribute, AttributeModifier> builder = ImmutableMultimap.builder();
+
+		Multimap<Attribute, AttributeModifier> defaultModifiers;
+
+		builder.putAll(ICurioItem.super.getAttributeModifiers(slotContext, uuid, stack));
+			builder.put(Attributes.ARMOR, new AttributeModifier(uuid, "Mistcloack modifier", 2, AttributeModifier.Operation.ADDITION));
```
leafreynolds: typo in mistcloak, though it doesn't super matter for this field

src/allomancy/java/leaf/cosmere/allomancy/common/items/MistcloakItem.java
```diff
@@ -29,7 +40,27 @@ public final String getArmorTexture(ItemStack stack, Entity entity, EquipmentSlo
...
+			builder.put(ForgeMod.ENTITY_GRAVITY.get(), new AttributeModifier(uuid, "Mistcloak glide", -0.02, AttributeModifier.Operation.ADDITION));
+			builder.put(Attributes.FLYING_SPEED, new AttributeModifier(uuid, "Mistcloack fly", 1.02, AttributeModifier.Operation.MULTIPLY_BASE));
```
leafreynolds: typo in mistcloak, though it doesn't super matter for this field

src/api/java/leaf/cosmere/api/helpers/CuriosHelper.java
```diff
@@ -0,0 +1,132 @@
...
+	public static List<SlotResult> getSlotsWithItem(LivingEntity entity, Item item)
+	{
+		ICuriosItemHandler sub = getCuriosHandler(entity);
+		if (sub == null)
+		{
+			return new ArrayList<SlotResult> ();
+		}
+		return sub.findCurios(item);
+	}
+	public static List<SlotResult> getSlotsWithItem(LivingEntity entity, Item... items)
```
leafreynolds, line 56: You call getSlotsWithItem 3 times here in a single function, and it does it for every item you're looping through.

call it once and assign it to a variable

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
```diff
@@ -110,5 +124,54 @@ public static void onLivingHurtEvent(LivingHurtEvent event)
...
+						stack.setDamageValue((int) (stack.getDamageValue() - absorbed));
+					}
+				}
+			}
+		}
+		// Find the curio item that acts as armor
+
 	}
+
+	private static ArrayList<ResourceKey<DamageType>> protectedDamageTypes = new ArrayList<>(
```
leafreynolds: Can be simplified to something like this, I think? 
```
private static final List<ResourceKey<DamageType>> protectedDamageTypes = 
List.of(
  DamageTypes.FALL,
  DamageTypes.FLY_INTO_WALL,
  DamageTypes.FREEZE
);
```

src/allomancy/java/leaf/cosmere/allomancy/client/render/renderer/MistcloakRenderer.java
```diff
@@ -52,9 +52,12 @@ public <T extends LivingEntity, M extends EntityModel<T>> void render(
 			return;
 		}
 
+
+
 		this.model.prepareMobModel(entity, limbSwing, limbSwingAmount, partialTicks);
 		this.model.setupAnim(entity, limbSwing, limbSwingAmount, ageInTicks, netHeadYaw, headPitch);
 
+
```
leafreynolds: A few extra lines added for no reason, lets remove these 🙂

src/datagen/allomancy/java/leaf/cosmere/allomancy/AllomancyCuriosProvider.java
```diff
@@ -0,0 +1,40 @@
...
+	{
+		super(Allomancy.MODID, output, fileHelper, registries);
+	}
+
+	@Override
+	public void generate(HolderLookup.Provider registries, ExistingFileHelper fileHelper)
+	{
+		//https://docs.illusivesoulworks.com/curios/slots/data-generation
+
+		this.createSlot("head")
```
leafreynolds, line 30: I think curios already generates a headslot by default, so we probably don't need to add this ourselves. needs double checking.

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
```diff
@@ -110,5 +124,54 @@ public static void onLivingHurtEvent(LivingHurtEvent event)
...
+			List.of(DamageTypes.FALL,
+					DamageTypes.FLY_INTO_WALL,
+					DamageTypes.FREEZE
```
leafreynolds: should we be absorbing freeze damage?

src/allomancy/java/leaf/cosmere/allomancy/common/items/MistcloakItem.java
leafreynolds: does this still get called, now that this doesn't extend from `ArmorItem` ?

src/allomancy/java/leaf/cosmere/allomancy/common/items/MistcloakItem.java
```diff
@@ -29,7 +40,27 @@ public final String getArmorTexture(ItemStack stack, Entity entity, EquipmentSlo
...
+	public boolean isDamageable(ItemStack stack)
+	{
+		return true;
+	}
+
+	public Multimap<Attribute, AttributeModifier> getAttributeModifiers(SlotContext slotContext, UUID uuid, ItemStack stack)
+	{
+		ImmutableMultimap.Builder<Attribute, AttributeModifier> builder = ImmutableMultimap.builder();
+
+		Multimap<Attribute, AttributeModifier> defaultModifiers;
```
leafreynolds: this var isn't needed, we can return builder.build() directly at the bottom

src/api/java/leaf/cosmere/api/helpers/CuriosHelper.java
```diff
@@ -0,0 +1,132 @@
...
+		ICuriosItemHandler sub = getCuriosHandler(entity);
+		assert sub != null;
+		return sub.findCurio(slotID, index).isPresent() ? sub.findCurio(slotID,index).get() : null;
+	}
+
+	public static ItemStack getStackInSlot(LivingEntity entity, String slotID, int index)
+	{
+		if(getSlot(entity, slotID, index) == null)
+		{
+			return null;
```
leafreynolds: should we be returning null here? usually we return an optional

src/api/java/leaf/cosmere/api/helpers/CuriosHelper.java
```diff
@@ -0,0 +1,132 @@
...
+		}
+		return Objects.requireNonNull(getSlot(entity, slotID, index)).stack();
+	}
+
+	public static SlotContext getContextInSlot(LivingEntity entity, String slotID, int index)
+	{
+		if(getSlot(entity, slotID, index) == null)
+		{
+			return null;
```
leafreynolds: should we be returning null here? usually we return an optional

leafreynolds (changes requested):

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
SoaringEaqle: I was thinking that if it's a wool cloak, it'll do something against the cold. I can remove it.

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/ShardplateCurioItem.java
SoaringEaqle: Curios creates a new itemstack when going into and out of a slot. The capability doesn't carry, so I create a new one and force the data over. I suppose I could try to figure out a way to make the capability transfer over instead, but this is working, so I'd rather not touch it.

src/datagen/allomancy/java/leaf/cosmere/allomancy/AllomancyCuriosProvider.java
SoaringEaqle: Not by default. The slot exists in the game, but we need to assign the player to have one in order for them to use it.

SoaringEaqle:
done

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/ShardbladeItem.java
```diff
@@ -1,3 +1,4 @@
+
```
leafreynolds: Still have a few files in the change list with random lines added, but I could probably go through and remove these? otherwise, the changes are looking good :)

leafreynolds:
@claude review

src/allomancy/java/leaf/cosmere/allomancy/client/render/model/MistcloakModel.java
```diff
@@ -12,6 +13,7 @@
 import com.mojang.blaze3d.vertex.PoseStack;
 import com.mojang.blaze3d.vertex.VertexConsumer;
 import leaf.cosmere.allomancy.common.Allomancy;
+import net.minecraft.client.model.HumanoidArmorModel;
```
leafreynolds: `HumanoidArmorModel` is imported but never used — can be removed.

src/allomancy/java/leaf/cosmere/allomancy/common/items/MistcloakItem.java
```diff
@@ -1,35 +1,57 @@
...
 
+import com.google.common.collect.ImmutableMultimap;
+import com.google.common.collect.Multimap;
 import leaf.cosmere.allomancy.common.Allomancy;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.EquipmentSlot;
+import net.minecraft.world.entity.ai.attributes.Attribute;
+import net.minecraft.world.entity.ai.attributes.AttributeModifier;
+import net.minecraft.world.entity.ai.attributes.Attributes;
 import net.minecraft.world.item.ArmorItem;
```
leafreynolds: Now that `MistcloakItem` extends `Item` instead of `ArmorItem`, these imports are unused and should be cleaned up:
- `ArmorItem` (line 16)
- `ArmorMaterial` (line 17)
- `Entity` (line 11)
- `EquipmentSlot` (line 12)
- `Nonnull` (line 24)

src/allomancy/java/leaf/cosmere/allomancy/common/items/MistcloakItem.java
```diff
@@ -1,35 +1,57 @@
...
 
-/* If we were to not use curios, this is what we would attempt to use.
+	public Multimap<Attribute, AttributeModifier> getAttributeModifiers(SlotContext slotContext, UUID uuid, ItemStack stack)
+	{
+		ImmutableMultimap.Builder<Attribute, AttributeModifier> builder = ImmutableMultimap.builder();
+
+		builder.putAll(ICurioItem.super.getAttributeModifiers(slotContext, uuid, stack));
+			builder.put(Attributes.ARMOR, new AttributeModifier(uuid, "Mistcloak modifier", 2, AttributeModifier.Operation.ADDITION));
+			builder.put(ForgeMod.ENTITY_GRAVITY.get(), new AttributeModifier(uuid, "Mistcloak glide", -0.02, AttributeModifier.Operation.ADDITION));
+			builder.put(Attributes.FLYING_SPEED, new AttributeModifier(uuid, "Mistcloak fly", 1.02, AttributeModifier.Operation.MULTIPLY_BASE));
```
leafreynolds: `MULTIPLY_BASE` with value `1.02` means the base flying speed gets an additional `1.02 * base` added, resulting in **+102%** flying speed — not +2%.

If a small boost is intended, use `0.02` instead. Or if the intent is to roughly double flying speed, then `1.0` with `MULTIPLY_BASE` would be clearer.

Also, the `builder.put` lines on 48-50 have an extra level of indentation compared to `builder.putAll` on line 47 and `return` on line 51 — looks like a formatting accident.

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
```diff
@@ -110,5 +124,54 @@ public static void onLivingHurtEvent(LivingHurtEvent event)
...
+						}
+
+						float original = event.getAmount();
+						float absorbed = original * 0.2f;
+						float remaining = original - absorbed;
+
+						// Reduce damage taken by entity
+						event.setAmount(remaining);
+
+						stack.setDamageValue((int) (stack.getDamageValue() + absorbed));
```
leafreynolds: Two issues here:

1. **No overflow protection** — if the item only has 5 durability left but `absorbed` is 10, `damageValue` will exceed `maxDamage`. Should clamp or break the item:
```java
int newDamage = Math.min(stack.getDamageValue() + Mth.ceil(absorbed), stack.getMaxDamage());
stack.setDamageValue(newDamage);
```

2. **Fractional truncation** — for small hits (e.g. 3 HP), `absorbed = 0.6f`, which truncates to `0` durability cost via the `(int)` cast. The item absorbs damage for free. Consider using `Mth.ceil()` instead.

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
```diff
@@ -110,5 +124,54 @@ public static void onLivingHurtEvent(LivingHurtEvent event)
...
+		if (entity.level().isClientSide || event.getAmount() <= 0)
+		{
+			return;
+		}
+		// Only do damage types that the mistcloak can protect.
+		for(ResourceKey<DamageType> type: protectedDamageTypes)
+		{
+			if(source.is(type))
+			{
+				if (CuriosHelper.getCuriosHandler(entity) != null)
```
leafreynolds: `getCuriosHandler()` returns `Optional<ICuriosItemHandler>`, not a nullable reference. This `!= null` check will always be true (an Optional is never null). Should be:
```java
if (CuriosHelper.getCuriosHandler(entity).isPresent())
```
Or better yet, skip this check entirely since `getSlotsWithItem` on the next line will just return an empty list if there's no handler.

src/api/java/leaf/cosmere/api/helpers/CuriosHelper.java
```diff
@@ -0,0 +1,138 @@
...
+
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.item.Item;
+import net.minecraft.world.item.ItemStack;
+import top.theillusivec4.curios.api.*;
+import top.theillusivec4.curios.api.type.capability.ICuriosItemHandler;
+
+import java.util.*;
+
+public class CuriosHelper
```
leafreynolds, line 11: Per project conventions, the `api` source set should remain dependency-free from external module implementations. This class depends directly on `CuriosApi` / `ICuriosItemHandler`. Consider moving it to `src/main/java/leaf/cosmere/common/helpers/` instead.

src/api/java/leaf/cosmere/api/CosmereTags.java
```diff
@@ -37,6 +38,7 @@ public static class Items
 		public static final TagKey<Item> CURIO_RING = makeItem(CuriosApi.MODID, "ring");
 		public static final TagKey<Item> CURIO_LEGS = makeItem(CuriosApi.MODID, "legs");
 		public static final TagKey<Item> CURIO_FEET = makeItem(CuriosApi.MODID, "feet");
+		public static final TagKey<Item> CURIO_SHARDPLATE = makeItem(CuriosApi.MODID, "shardplate");
```
leafreynolds: `CURIO_SHARDPLATE` appears to be part of a separate shardplate-as-curio feature. Consider splitting this into its own PR to keep this one focused on the mistcloak.

leafreynolds (commented):
## Mistcloak v2 Review

Good concept converting the Mistcloak from ArmorItem to a Curios ICurioItem. The model rework is a major simplification — dropping all those complex rotated sub-parts in favor of a cleaner root-based hierarchy is a big readability win.

### Bugs
- **Flying speed modifier** uses `MULTIPLY_BASE` with value `1.02`, which gives **+102% base flying speed**, not +2%. If a small boost was intended, the value should be `0.02`.
- **Durability overflow** — absorbed damage can push `damageValue` past `maxDamage` without the item breaking. Need a bounds check.
- Damage value semantics were already flagged by @leafreynolds — `damageValue` is damage *taken* (0 = pristine), so the guard and arithmetic need adjusting.

### Cleanup Needed
- Several now-unused imports after the `ArmorItem` → `Item` conversion (`ArmorItem`, `ArmorMaterial`, `Entity`, `EquipmentSlot`, `Nonnull`)
- `CuriosHelper` is placed in the `api` source set but depends on CuriosApi — per project conventions, `api` should remain dependency-free from external module implementations
- `CURIO_SHARDPLATE` tag addition and multiple surgebinding whitespace-only changes seem unrelated to this PR's scope
- Unused `ForgeMod` import in `AllomancyRecipeGen`

### Missing
- I don't see a curio slot tag entry (e.g. `data/curios/tags/items/head.json`) that registers the mistcloak item for the head slot. Without this, Curios won't know the item can be equipped there. Worth verifying this works in-game.

See inline comments for specifics.

src/datagen/allomancy/java/leaf/cosmere/allomancy/AllomancyRecipeGen.java
```diff
@@ -12,6 +12,7 @@
 import net.minecraft.data.recipes.*;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.world.item.Items;
+import net.minecraftforge.common.ForgeMod;
```
leafreynolds: Unused import — `ForgeMod` isn't referenced anywhere in this file.

src/api/java/leaf/cosmere/api/helpers/CuriosHelper.java
leafreynolds: ignore

src/allomancy/java/leaf/cosmere/allomancy/common/items/MistcloakItem.java
Gerbagel: Is this even true? Claude doesn't cite any documentation of anything. Probably something to check, at least. Worst case it just hallucinated, as usual.

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
Gerbagel: The item breaking might be handled somewhere else. Again, something to check.

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
Gerbagel: Already handled.

src/api/java/leaf/cosmere/api/CosmereTags.java
Gerbagel: Ignore. This is true to keep in mind for future PRs, but it's really not so important.

src/datagen/allomancy/java/leaf/cosmere/allomancy/AllomancyRecipeGen.java
Gerbagel: I'm not gonna check every unused import. If they're removed, cool. If not, no huge loss, someone'll probably catch them later.

src/allomancy/java/leaf/cosmere/allomancy/common/items/MistcloakItem.java
SoaringEaqle: It is seemingly right about this.

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
SoaringEaqle: I went ahead and forced it to happen any way, plus did fix the math up a bit.

PR #188, opened 2025-11-14
Improved lavis polyp

leeislate:
I updated the lavis polyp model and texture to look a little more minecrafty.


leafreynolds:
Hey Lee, did you test this ? I think we have the polyp in game already, but it looks like you added files instead of replacing the ones already there

leeislate:
Oh, I thought I removed the old ones.

leafreynolds:
I think what happened was that the original has an underscore, so it didn't get overwritten by your new one.

Do you want to fix it? or should I 😄

leeislate:
I’d love it if you could… I’m not sure how…

PR #193, opened 2025-11-16
Shardplate and Shardblade stuff

SoaringEaqle:

See changes in mistcloak_v2 pull request relating to shardplate


src/allomancy/java/leaf/cosmere/allomancy/common/items/MistcloakItem.java
```diff
@@ -62,4 +62,4 @@ public HumanoidModel<?> getHumanoidArmorModel(LivingEntity livingEntity, ItemSta
 			}
 		});
 	}*/
-}
+}
```
leafreynolds: I wonder if we can make the pull request realise there are no changes here, rather than a change to end of file

src/allomancy/java/leaf/cosmere/allomancy/common/eventHandlers/AllomancyEntityEventHandler.java
```diff
@@ -111,4 +98,4 @@ public static void onLivingHurtEvent(LivingHurtEvent event)
 		AllomancyPewter.onLivingHurtEvent(event);
 		AllomancyChromium.onLivingHurtEvent(event);
 	}
-}
+}
```
leafreynolds: I wonder if we can make the pull request realise there are no changes here, rather than a change to end of file

src/allomancy/resources/assets/allomancy/textures/models/armor/mistcloak.png
leafreynolds, line 1: we should try zero out these file changes, since it might conflict with your other pull request

src/allomancy/resources/assets/allomancy/textures/item/mistcloak.png
leafreynolds, line 1: we should try zero out these file changes, since it might conflict with your other pull request

src/api/java/leaf/cosmere/api/helpers/CuriosHelper.java
leafreynolds, line 1: this file will probably conflict with your other pull request, so will need to be careful when we merge them both.

See other commit for review comments on this file.

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/DynamicShardplateData.java
```diff
@@ -0,0 +1,215 @@
...
+		return this.nbt;
+	}
+
+	@Override
+	public void deserializeNBT(CompoundTag compoundTag)
+	{
+		super.deserializeNBT(compoundTag);
+
+
+		this.headID = nbt.getString("headID");
```
leafreynolds: a bunch of these are read in more than once

headID, bodyID, rightArmID, rightLegID, and kamaID

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/DynamicShardplateData.java
```diff
@@ -0,0 +1,215 @@
...
+		this.rightArmID = nbt.getString("rightArmID");
+		this.bodyID = nbt.getString("bodyID");
+		this.rightLegID = nbt.getString("rightLegID");
+		this.kamaID = nbt.getString("kamaID");
+
+		this.headID = nbt.getString("headID");
```
leafreynolds, line 91: a bunch of these are read in more than once

headID, bodyID, rightArmID, rightLegID, and kamaID

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/DynamicShardplateData.java
leafreynolds: this top section is probably an accident

src/surgebinding/java/leaf/cosmere/surgebinding/client/render/model/DynamicShardplateModel.java
```diff
@@ -0,0 +1,513 @@
...
+	public static final int TOTAL_HELMET_IDS = 1;
+	public static final int TOTAL_FACEPLATE_IDS = 4;
+	public static final int TOTAL_ARM_IDS = 1;
+	public static final int TOTAL_PALDRON_IDS = 2;
+	public static final int TOTAL_TORSO_IDS = 1;
+	public static final int TOTAL_LEG_IDS = 1;
+	public static final int TOTAL_KAMA_IDS = 1;
+	public static final int TOTAL_BOOT_IDS = 1;
+
+	public static final ModelLayerLocation LAYER_LOCATION = new ModelLayerLocation(new ResourceLocation("modid", "shardplate"), "main");
```
leafreynolds: "modid" needs to be replaced with our modid
` Surgebinding.rl(...)`

src/surgebinding/java/leaf/cosmere/surgebinding/client/render/renderer/ArmorRenderer.java
```diff
@@ -57,7 +67,13 @@ public <T extends LivingEntity, M extends EntityModel<T>> void render(
 
 		ICurioRenderer.followBodyRotations(entity, this.model);
 
+		this.model.render(stack, slotContext, matrixStack, renderTypeBuffer, light);
 		//this.model.render(stack, slotContext, matrixStack, renderTypeBuffer, light);
```
leafreynolds: did you mean to add a new line identical to the commented out version? should just uncomment the code

src/surgebinding/java/leaf/cosmere/surgebinding/client/render/model/DynamicShardplateModel.java
```diff
@@ -0,0 +1,513 @@
...
+			VertexConsumer lightLayer = buffer.getBuffer(RenderType.armorCutoutNoCull(TRIM));
+			renderToBuffer(matrixStack,
+					lightLayer,
+					LightTexture.FULL_BRIGHT,
+					OverlayTexture.NO_OVERLAY,
+					glyphColor.getRed()/255F,
+					glyphColor.getGreen()/255F,
+					glyphColor.getBlue()/255F,
+					1F);
+*/
```
leafreynolds, line 433: should probably remove sections of commented out code like this if we're not going to use them.

src/main/resources/data/cosmere/loot_tables/inject/ancient_city.json
```diff
@@ -0,0 +1,56 @@
+{
+  "__comment": "This table is injected as its own pool into minecraft:chests/woodland_mansion",
```
leafreynolds: comment should probably say ancient_city

src/main/resources/data/cosmere/loot_tables/inject/stronghold_corridor.json
```diff
@@ -19,6 +19,17 @@
             }
           ]
         },
+        {
+
+          "type": "minecraft:item",
+          "_comment": "Shardblade is randomized.",
+          "name": "surgebinding:shardblade",
```
leafreynolds, line 26: 'weight' is missing from a couple of these loot tables. Just looked up what that does (I know I leave it out sometimes too), and apparently that means it defaults to 1.

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/ShardData.java
```diff
@@ -0,0 +1,184 @@
...
+		this.stack = stack;
+		this.nbt = new CompoundTag();
+	}
+
+	public ShardData(ItemStack stack, int i)
+	{
+		this.stack = stack;
+		this.nbt = new CompoundTag();
+
+		if(i == 1)
```
leafreynolds: magic number of 1 could probably be replaced.

If all we're checking is a boolean state, replace i with a bool.

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/SurgebindingSpiritwebSubmodule.java
```diff
@@ -114,6 +117,8 @@ public void tickServer(ISpiritweb spiritweb)
 			if (stormlightStored > 0 && surgebindingActiveTick)
 			{
 				//being hurt takes priority
+
```
leafreynolds: space for no reason

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/SurgebindingSpiritwebSubmodule.java
```diff
@@ -128,6 +133,8 @@ public void tickServer(ISpiritweb spiritweb)
 				//otherwise conditional effects
 				else
 				{
+					List<SlotResult> plate = CuriosHelper.getSlotsByIdentifier(livingEntity, "shardplate");
```
leafreynolds: new var that never gets referenced. should be removed

src/datagen/surgebinding/java/leaf/cosmere/surgebinding/SurgebindingCuriosProvider.java
```diff
@@ -0,0 +1,41 @@
...
+				.size(1)
+				.dropRule(ICurio.DropRule.ALWAYS_KEEP)
+				.addCosmetic(true);
+
+		this.createEntities("shardbearers")
+				.addPlayer()
+				.addEntities(EntityType.ARMOR_STAND, EntityType.VINDICATOR, EntityType.PILLAGER, EntityType.VILLAGER)
+				.addSlots("shardplate");
+	}
+}
```
leafreynolds: I think it's complaining here because it's missing a new line at the end of the file, which it's doing on several files throughout the pull request. I'll see about updating our default editor config file to default that to adding the new line eof.

src/surgebinding/java/leaf/cosmere/surgebinding/client/SurgebindingModClientEvents.java
```diff
@@ -36,13 +43,38 @@ public static void registerLayers(final EntityRenderersEvent.RegisterLayerDefini
...
+	@SubscribeEvent
+	public static void registerItemColors(RegisterColorHandlersEvent.Item event)
+	{
+		event.register((stack, tintIndex) -> {
+			// We only want to tint layer1 (index 1)
+			if (tintIndex != 0) return Color.WHITE.getRGB(); // white = no tint
+
+			return (stack.getCapability(ShardData.SHARD_DATA)
+					.map(cap -> {
+						DynamicShardplateData cap2 = (DynamicShardplateData)cap;
```
leafreynolds: let's replace this cast with an `instanceof ` check

src/api/java/leaf/cosmere/api/Roshar.java
```diff
@@ -305,5 +308,126 @@ public Color getColor()
...
+		public Gemstone getGemstone()
+		{
+			return gemstone;
+		}
+
+		public boolean hasBlade()
+		{
+			switch (this)
+			{
+				case BONDSMITH:
```
leafreynolds: for small switch statements like this, you're better off replacing it with an if/else statement.

src/api/java/leaf/cosmere/api/Roshar.java
```diff
@@ -305,5 +308,126 @@ public Color getColor()
...
+				case TRUTHWATCHER -> "mistspren";
+				case LIGHTWEAVER -> "cryptic";
+				case ELSECALLER -> "inkspren";
+				case WILLSHAPER -> "lightspren";
+				case STONEWARD -> "peakspren";
+				case BONDSMITH -> null;
+
+			};
+		}
+		public String getSprenOrBondsmith(int spren)
```
leafreynolds, line 360: I think these get spren or bondsmith functions could do with a bit of clean up. Not strictly necessary

src/datagen/allomancy/java/leaf/cosmere/allomancy/AllomancyDataGenerator.java
```diff
@@ -1,11 +1,15 @@
 /*
  * File updated ~ 8 - 10 - 2024 ~ Leaf
+
```
leafreynolds: File has a bunch of unneeded blank line changes

src/datagen/surgebinding/java/leaf/cosmere/surgebinding/SurgebindingDataGenerator.java
```diff
@@ -1,5 +1,7 @@
+
```
leafreynolds: File has a bunch of unneeded blank line changes

src/datagen/surgebinding/java/leaf/cosmere/surgebinding/SurgebindingTagsProvider.java
```diff
@@ -1,3 +1,4 @@
+
```
leafreynolds, line 1: File has a bunch of unneeded blank line changes

src/main/java/leaf/cosmere/common/properties/PropTypes.java
```diff
@@ -42,8 +42,18 @@ public static class Items
...
 
+		public static final Supplier<Item.Properties> SHARDPLATE = () -> new Item.Properties()
+				.stacksTo(1)
+				.rarity(Rarity.EPIC)
+				.fireResistant();
+
 		public static final Supplier<Item.Properties> ONE = () -> new Item.Properties().stacksTo(1);
 		public static final Supplier<Item.Properties> SIXTEEN = () -> new Item.Properties().stacksTo(16);
 		public static final Supplier<Item.Properties> SIXTY_FOUR = () -> new Item.Properties().stacksTo(64);
+		public static final Supplier<Item.Properties> MISTCLOAK = () -> new Item.Properties()
```
leafreynolds, line 53: remove, as it's part of other pull request

leafreynolds (changes requested):

src/api/java/leaf/cosmere/api/helpers/CuriosHelper.java
```diff
@@ -0,0 +1,132 @@
...
+public class CuriosHelper
+{
+	public static ICuriosItemHandler getCuriosHandler(LivingEntity entity)
+	{
+		return CuriosApi.getCuriosInventory(entity).resolve().isPresent()? CuriosApi.getCuriosInventory(entity).resolve().get(): null;
+	}
+
+	public static SlotResult getSlot(LivingEntity entity, String slotID, int index) {
+		ICuriosItemHandler sub = getCuriosHandler(entity);
+		assert sub != null;
```
leafreynolds: `assert` is disabled by default in production JVMs (`-ea` must be explicitly passed). This will throw a `NullPointerException` instead of an `AssertionError` when `getCuriosHandler` returns null. Replace with an explicit null check or change the return type to `Optional`.

src/api/java/leaf/cosmere/api/helpers/CuriosHelper.java
```diff
@@ -0,0 +1,132 @@
...
+		return sub.findCurio(slotID, index).isPresent() ? sub.findCurio(slotID,index).get() : null;
+	}
+
+	public static ItemStack getStackInSlot(LivingEntity entity, String slotID, int index)
+	{
+		if(getSlot(entity, slotID, index) == null)
+		{
+			return null;
+		}
+		return Objects.requireNonNull(getSlot(entity, slotID, index)).stack();
```
leafreynolds: `getSlot` is called twice here — once to null-check and again inside `Objects.requireNonNull`. Cache the result in a local variable to avoid the double lookup.

leafreynolds (commented):
Shardplate / Shardblade refactor review — the consolidation to a single Curios slot item is the right architectural direction. A few bugs and cleanup items noted inline.

src/api/java/leaf/cosmere/api/helpers/CuriosHelper.java
leafreynolds: `assert` is disabled by default in production JVMs (`-ea` must be explicitly passed). This will silently become a `NullPointerException` when `getCuriosHandler` returns null. Replace with an explicit null check or return an `Optional<SlotResult>`.

src/api/java/leaf/cosmere/api/helpers/CuriosHelper.java
leafreynolds: `getSlot` is called twice here — once for the null check and again inside `Objects.requireNonNull`. Cache the result:
```java
SlotResult slot = getSlot(entity, slotID, index);
if (slot == null) return null;
return slot.stack();
```

src/surgebinding/java/leaf/cosmere/surgebinding/client/render/model/DynamicShardplateModel.java
leafreynolds: Leftover Blockbench export template — the namespace `"modid"` is a placeholder, not the actual mod ID. This field is never referenced so it does not crash, but if anyone ever uses it the ResourceLocation will be wrong. Either remove `LAYER_LOCATION` or fix it to `Surgebinding.rl("shardplate")`.

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/DynamicShardplateData.java
leafreynolds, line 91: Duplicate assignments — `headID`, `bodyID`, `rightArmID`, `rightLegID`, and `kamaID` are each read twice from NBT. Lines 116–120 are dead code and should be deleted.

src/surgebinding/java/leaf/cosmere/surgebinding/client/SurgebindingModClientEvents.java
leafreynolds: Unchecked cast — `cap` is typed as `IShard`, but a `ShardbladeItem` would also satisfy `ShardData.SHARD_DATA` and its capability is also `IShard`. If a Shardblade stack were ever passed to this color handler, this cast would throw a `ClassCastException`. Add an `instanceof` guard:
```java
if (!(cap instanceof DynamicShardplateData cap2)) return Roshar.getDeadplate().getRGB();
```

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/SurgebindingSpiritwebSubmodule.java
leafreynolds: Dead code — `plate` is fetched but never used. This runs every tick for any surgebinding player. Either use it or remove the line.

src/main/resources/data/cosmere/loot_tables/inject/ancient_city.json
leafreynolds: Copy-paste error in comment — says `woodland_mansion` but this file is `ancient_city.json`.

SoaringEaqle:
Done

leafreynolds:
@claude review

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/IShardItem.java
```diff
@@ -0,0 +1,91 @@
...
+		return getShardData(stack).isLiving();
+	}
+
+	default LivingEntity getBond(ItemStack stack)
+	{
+		ShardData data = getShardData(stack);
+		UUID id = data.getBondedEntity();
+		if(id != null)
+		{
+			Level level = Minecraft.getInstance().level;
```
leafreynolds: **Bug (server crash):** `Minecraft.getInstance().level` is a client-only class. This code is in common (non-sided) code and runs during `inventoryTick` / `curioTick` on the server. This will crash a dedicated server with `NoClassDefFoundError`.

This needs to be reworked to take a `Level` parameter from the caller, or the bond lookup needs to happen server-side via `ServerLevel.getPlayerByUUID()`.

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/IShardItem.java
```diff
@@ -0,0 +1,91 @@
...
+
+		}
+		else
+		{
+			return null;
+		}
+	}
+
+	default boolean isBondedTo(ItemStack stack, LivingEntity player)
+	{
```
leafreynolds: **Bug (NPE):** `getBond(stack)` can return `null` (when there's no bond, or `UNKEYED_UUID`), but `.is(player)` is called on the result without null-checking. Should be:
```java
LivingEntity bond = getBond(stack);
return bond != null && bond.is(player);
```

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/ShardbladeItem.java
```diff
@@ -54,9 +74,101 @@ public boolean isFoil(ItemStack itemStack)
...
+	@Override
+	public void bond(ItemStack stack, Player entity)
+	{
+		// if bonded, then don't bond again
+		if (getBond(stack) != null)
+		{
+			return;
+		}
+
+
```
leafreynolds: **Bug (NPE):** Same issue as `IShardItem.isBondedTo()` — `getBond(stack)` can return null here. Needs a null check before calling `.is(player)`.

src/surgebinding/java/leaf/cosmere/surgebinding/common/network/packets/SyncShardCapMessage.java
```diff
@@ -0,0 +1,54 @@
...
+				stack.getCapability(ShardData.SHARD_DATA)
+						.ifPresent(cap -> cap.deserializeNBT(nbt));
+			}
+		});
+		context.setPacketHandled(true);
+	}
+
+	@Override
+	public void encode(FriendlyByteBuf buffer)
+	{
```
leafreynolds, line 49: **Bug (broken packet):** `encode()` is empty but `decode()` reads `readNbt()` and `readInt()`. This packet will never work — the server will write nothing to the buffer, but the client will try to read NBT + int from it. Needs:
```java
buffer.writeNbt(this.nbt);
buffer.writeInt(this.slot);
```

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/ideals/RadiantStateManager.java
```diff
@@ -271,6 +283,24 @@ private void onSuccessfulIdealSworn(@NotNull SpiritwebCapability spiritweb)
 					player.sendSystemMessage(Component.literal("THESE WORDS ARE ACCEPTED."));
 					updatePowerState();
 				}
+				if(ideal == 3)
+				{
+					blade = new ItemStack(SurgebindingItems.SHARDBLADE.asItem());
+					CompoundTag tag = plate.getOrCreateTag();
```
leafreynolds: **Bug (NPE):** When `ideal == 3`, this creates `blade` but then calls `plate.getOrCreateTag()`. At this point `plate` is still null — it's only assigned when `ideal == 4`. This should be `blade.getOrCreateTag()`.

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/ideals/RadiantStateManager.java
```diff
@@ -309,4 +339,63 @@ private void updatePowerState()
...
+
+	// For use with commands. Possible bondsmith applications?
+	public void forceSwear(int idealToSwear, Roshar.RadiantOrder idealOrder)
+	{
+		this.ideal = idealToSwear;
+		onSuccessfulIdealSworn(spiritweb);
+		if(ideal >= 3 && blade == null)
+		{
+			blade = new ItemStack(SurgebindingItems.SHARDBLADE.asItem());
+			CompoundTag tag = plate.getOrCreateTag();
```
leafreynolds: **Bug (NPE):** Same issue in `forceSwear()` — when `ideal >= 3 && blade == null`, creates blade then calls `plate.getOrCreateTag()`, but plate could still be null if ideal is exactly 3. Should be `blade.getOrCreateTag()`.

src/surgebinding/java/leaf/cosmere/surgebinding/common/items/ShardplateCurioItem.java
```diff
@@ -0,0 +1,406 @@
...
+				ssm.adjustStormlight(-charge, true);
+			}
+		}
+		if(((ShardplateCurioItem)stack.getItem()).getCharge(stack) != 0)
+		{
+			entity.addEffect(EffectsHelper.getNewEffect(MobEffects.JUMP, 1));
+		}
+		if(((ShardplateCurioItem)stack.getItem()).getCharge(stack) <= ((ShardplateCurioItem)stack.getItem()).getMaxCharge(stack)/4)
+		{
+			ParticleHelper.spawnLeakEffect((ServerLevel) entity.level(), 2, entity);
```
leafreynolds: **Bug (client crash):** `curioTick` runs on both client and server. The cast `(ServerLevel) entity.level()` will throw `ClassCastException` on the client. This whole block (and the jump effect above it) needs a `if (!entity.level().isClientSide)` guard, or at minimum the `ParticleHelper.spawnLeakEffect` call does.

src/surgebinding/java/leaf/cosmere/surgebinding/common/recipes/ShardplateChargingRecipe.java
```diff
@@ -0,0 +1,253 @@
...
+	public boolean canCraftInDimensions(int width, int height)
+	{
+		//if you can fit 2 items, the bottle and a metal, you can combine
+		return width * height > 1;
+	}
+
+	@Override
+	public @Nonnull ResourceLocation getId()
+	{
+		return Surgebinding.rl("vial_mix");
```
leafreynolds: Copy-paste leftover — `getId()` returns `"vial_mix"` but should be something like `"plate_charging"` to match the actual recipe.

leafreynolds (commented):
## Shardplate / Shardblade Rework Review

The consolidation from 4-piece vanilla ArmorItem to a single Curios-based ShardplateCurioItem is the right direction. The unified `ShardData` / `IShard` capability shared between blades and plate is clean. Several crash-level bugs need fixing before merge — details in inline comments.

### Critical (will crash)
- `IShardItem.getBond()` references `Minecraft.getInstance()` from common code — dedicated server crash
- `IShardItem.isBondedTo()` and `ShardbladeItem.canSummonDismiss()` — NPE when `getBond()` returns null
- `SyncShardCapMessage.encode()` is empty — packet will fail to send/receive
- `RadiantStateManager.onSuccessfulIdealSworn()` at ideal 3 — calls `plate.getOrCreateTag()` but `plate` is null (only created at ideal 4). Same bug in `forceSwear()`
- `ShardplateCurioItem.curioTick()` casts to `ServerLevel` without a client-side guard — client crash

### Moderate
- `ShardplateChargingRecipe.getId()` returns `"vial_mix"` — copy-paste from allomancy recipe
- `SetupShardDataLootFunction` uses `new Random()` instead of `pContext.getRandom()`
- `registerItemColors` tint index logic may be inverted (tintIndex 0 = layer0, not layer1)
- Double capability registration: both `initCapabilities()` and `AttachCapabilitiesEvent<ItemStack>` attach `SHARD_DATA` — could conflict
- `DynamicShardplateData` constructor uses `new String(x).replace()` — unnecessary, just use `x.replace()`

### Minor
- `.DS_Store` should be in `.gitignore`, not tracked
- Several files have `new String()` wrappers that aren't needed (strings are immutable in Java)
- `ShardplateArmorMaterial` no longer implements `ArmorMaterial` — may be dead code now that `ShardplateItem` is deleted

src/surgebinding/java/leaf/cosmere/surgebinding/common/loot/SetupShardDataLootFunction.java
```diff
@@ -0,0 +1,81 @@
...
+		super(pPredicates);
+	}
+
+	@Override
+	protected ItemStack run(ItemStack pStack, LootContext pContext)
+	{
+		if(pStack.getItem() instanceof IShardItem shard)
+		{
+			Random rand = new Random();
+			int chance = rand.nextInt(0, 40);
```
leafreynolds: Should use `pContext.getRandom()` instead of `new Random()` — loot functions should use the context's random for determinism (e.g. structure seeds, replays).

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/DynamicShardplateData.java
```diff
@@ -0,0 +1,208 @@
...
+		this.rightPaldronsID = "right_paldron" + MathHelper.randomInt(0, DynamicShardplateModel.TOTAL_PALDRON_IDS);
+		this.rightLegID = "rightleg_top" + MathHelper.randomInt(1, DynamicShardplateModel.TOTAL_LEG_IDS);
+		this.rightBootOutsideID = "rightboot_outside" + MathHelper.randomInt(1, DynamicShardplateModel.TOTAL_BOOT_IDS);
+		this.rightBootTipID = new String(rightBootOutsideID).replace("outside","tip");
+
+		this.leftArmID = new String(rightArmID).replace("right","left");
+		this.leftPaldronsID = new String(rightPaldronsID).replace("right","left");
+		this.leftLegID = new String(rightLegID).replace("right","left");
+		this.leftBootOutsideID = new String(rightBootOutsideID).replace("right","left");
+		this.leftBootTipID = new String(rightBootTipID).replace("right","left");
```
leafreynolds: `new String(rightBootOutsideID).replace(...)` — the `new String()` copy is unnecessary in Java. Strings are immutable, so `.replace()` already returns a new String. Just use `rightBootOutsideID.replace("outside", "tip")` directly. Same applies to the left-side mirroring lines below.

src/surgebinding/java/leaf/cosmere/surgebinding/client/SurgebindingModClientEvents.java
```diff
@@ -36,13 +43,42 @@ public static void registerLayers(final EntityRenderersEvent.RegisterLayerDefini
 		SurgebindingLayerDefinitions.register(evt);
 	}
 
+	@SubscribeEvent
+	public static void registerItemColors(RegisterColorHandlersEvent.Item event)
+	{
+		event.register((stack, tintIndex) -> {
+			// We only want to tint layer1 (index 0)
+			if (tintIndex != 0) return Color.WHITE.getRGB(); // white = no tint
```
leafreynolds: The comment says "We only want to tint layer1 (index 0)" but tintIndex 0 corresponds to `layer0`, not `layer1`. As written, this returns white (no tint) for layer0 and applies the order color to layer1 — double-check this is the intended behavior, since the item model has `layer0: shardplate_helmet` and `layer1: shardplate_helmet_visor`.

leafreynolds:
side note after the claude review, did you get a chance to test this in a server environment? via runServer and have runClient connect to it?

Issue #77, opened 2023-12-06
[Feruchemy] Nicrosil


leafreynolds:
# How it currently works:
Currently in the mod, Feruchemical Nicrosil does not match canon. I originally implemented it to work similarly to the other powers. Store for x time to tap for x time. In this case, store your ability to use your powers, to be able to tap and double your strength in your powers. 

[This does not match canon](https://wob.coppermind.net/events/316/#e11247). 

Additionally, though more a bug than an intended feature, only someone who has f-nicrosil as a power can tap a nicrosilmind, where [anyone is supposed to be allowed to tap nicrosil](https://wob.coppermind.net/events/4/#e5176) (but not store).

# Canon
[Brandon has confirmed that the f-nicrosil portion of the medallion acts similarly to a coppermind, where you have one copy of the thing that you can gain and can return](https://wob.coppermind.net/events/316/#e11247). This could be a specific function of the medallion, which is more restrictive than an unbound nicrosil ferring. 

# Proposed Changes
I would like to shift things to be closer towards canon. In that respect, F-nicrosil should be able to store all investiture related information of the player's spiritweb. This includes things like powers, breaths, stormlight, shardblades etc.

While stored, the player will not be able to use whatever they stored until they tap it back. This means players could conceivably lose the metalmind forever. This is acceptable.

The main question is: How do we allow this? What is the process in-game that a player will need to go through in order to store something. They will need to be allowed to be precise, storing only what they want to store.

I have several ideas that could give players precision, of varying complexity:
- Commands, something like `/feruchemy nicrosil store breaths` or  `/feruchemy nicrosil store power allomancy:pewter`
- An awakening style command, activated by chat like `"My life to yours, my power become yours."`. Exact wording unknown. Will need to have the ability to hide the chat from others.
- Full blown dynamic UI, showing you what you have access to in your list, and possible places to store it. Ability to see what you have stored in metalminds in your inventory. If the metalmind is keyed, you can see it is keyed, but not what is within it.

AngelOfCreation:
+1 to the UI idea. Question: will players be able to store powers gained from spikes in nicrosil minds? If not, WILL there be a way to store all the powers in a mind and create the bands of mourning, give it away, and then get them back?

leafreynolds:
Interesting question. I would suggest not giving your powers away.

leafreynolds:
[...]

PR #196, opened 2025-11-24
Feruchemical Nicrosil Improvements

RequestingKidney:
## Changes proposed in this pull request (partially written by copilot...)
This pull request introduces significant improvements to the handling of powers regarding Nicrosil in Feruchemy. It adds new item classes for Nicrosil bracelet and necklace metalminds, refactors how powers are managed and queried via attributes, and improves the user interface for displaying Nicrosil investiture. Additionally, it removes duplicated tooltip logic from bracelet and necklace items, centralizing it in the new classes.

**Enhancements to power handling and investiture:**

* Introduced new classes like `NicrosilBraceletMetalmindItem`  to represent Nicrosil-specific bracelet and necklace metalminds, inheriting from `PowerMetalCurioItem` and providing custom capacity and tooltip handling.
* Refactored the logic for giving starting items so that Nicrosil rings are handled as generic `Item` rather than as `RingMetalmindItem`.
* Removed redundant tooltip code for Nicrosil investiture from both `BraceletMetalmindItem` and `NecklaceMetalmindItem`, as this logic is now handled in the new Nicrosil item classes.

**Attribute and power querying improvements:**

* Added `getPowers()` methods to the `ISpiritwebSubmodule` interface to provide default implementations for querying powers and entity power instances. 

**User interface and data generation updates:**

* Added a GUI overlay to render Nicrosil HUD information, and ensured the HUD is updated based on the player's current powers.
* Updated item model and language data generators to recognize and handle `PowerMetalCurioItem` (the new base for Nicrosil metalminds), ensuring correct tooltips and models are generated.
**Other minor improvements:**

* Improved code clarity and maintainability by updating imports, fixing comments, and clarifying parameter names (e.g., in `MathHelper.InverseLerp`).


RequestingKidney:
@leafreynolds This PR is officially ready for review. Please link it to https://github.com/leafreynolds/cosmere/issues/77.

RequestingKidney:
Link to demo: https://discord.com/channels/828383636465123328/837601788755836948/1442981148631498953

src/api/java/leaf/cosmere/api/Manifestations.java
```diff
@@ -15,11 +15,11 @@ public class Manifestations
 {
 	public final static int ALLOMANCY_ID = 1;
 	public final static int FERUCHEMY_ID = 2;
-	public final static int HEMALURGY_ID = 7;//yes I know. Didn't add this till way later and now I'm too afraid to change it... todo: change it
-	public final static int SURGEBINDING_ID = 3;
-	public final static int AONDOR_ID = 4;
-	public final static int AWAKENING_ID = 5;
-	public final static int SANDMASTERY_ID = 6;
+	public final static int HEMALURGY_ID = 3;//yes I know. Didn't add this till way later and now I'm too afraid to change it... I did what needed to be done
```
leafreynolds: pls no

src/feruchemy/java/leaf/cosmere/feruchemy/common/registries/FeruchemyItems.java
```diff
@@ -50,11 +49,24 @@ public class FeruchemyItems
 	public static final Map<Metals.MetalType, ItemRegistryObject<NecklaceMetalmindItem>> METAL_NECKLACES =
 			Arrays.stream(EnumUtils.METAL_TYPES)
 					.filter(Metals.MetalType::hasFeruchemicalEffect)
+					.filter(metalType -> metalType != Metals.MetalType.NICROSIL)
 					.collect(Collectors.toMap(
 							Function.identity(),
 							type -> ITEMS.register(
 									type.getName() + RegNameStubs.NECKLACE + RegNameStubs.METALMIND,
 									() -> new NecklaceMetalmindItem(type)
```
leafreynolds: we shouldn't need to treat nicrosil as a separate list of items. You can just switch here if metal type is nicrosil and use the different class

src/datagen/feruchemy/java/leaf/cosmere/feruchemy/FeruchemyTagProvider.java
```diff
@@ -35,10 +35,19 @@ protected void registerTags(HolderLookup.Provider registries)
...
 			{
-				//curio stuff
-				getItemBuilder(CosmereTags.Items.CURIO_NECKLACE).add(FeruchemyItems.METAL_NECKLACES.get(metalType).asItem());
-				getItemBuilder(CosmereTags.Items.CURIO_RING).add(FeruchemyItems.METAL_RINGS.get(metalType).asItem());
-				getItemBuilder(CosmereTags.Items.CURIO_BRACELET).add(FeruchemyItems.METAL_BRACELETS.get(metalType).asItem());
+				// Using switch for future metal types that may need special handling
+				switch(metalType)
+				{
+					case NICROSIL:
+						getItemBuilder(CosmereTags.Items.CURIO_NECKLACE).add(FeruchemyItems.NICROSIL_METAL_NECKLACE.asItem());
```
leafreynolds: which would mean all these extra places can use their original code, rather than a new list for no reason

src/datagen/feruchemy/java/leaf/cosmere/feruchemy/FeruchemyRecipeGen.java
```diff
@@ -49,9 +50,19 @@ protected static void addBasicMetalmindRecipes(Consumer<FinishedRecipe> consumer
...
 		if (metalType.hasFeruchemicalEffect())
 		{
-			ShapedRecipeBuilder.shaped(RecipeCategory.TOOLS, FeruchemyItems.METAL_NECKLACES.get(metalType)).define('X', inputMaterial).pattern("XXX").pattern("X X").pattern(" X ").group("necklace").unlockedBy("has_material", has(inputMaterial)).save(consumer);
-			ShapedRecipeBuilder.shaped(RecipeCategory.TOOLS, FeruchemyItems.METAL_RINGS.get(metalType)).define('X', inputMaterial).pattern(" X ").pattern("X X").pattern(" X ").group("ring").unlockedBy("has_material", has(inputMaterial)).save(consumer);
-			ShapedRecipeBuilder.shaped(RecipeCategory.TOOLS, FeruchemyItems.METAL_BRACELETS.get(metalType)).define('X', inputMaterial).pattern(" X ").pattern("X X").pattern("X X").group("bracelet").unlockedBy("has_material", has(inputMaterial)).save(consumer);
+			// Using switch for future metal types that may need special handling
+			switch(metalType)
+			{
+				case NICROSIL:
+					ShapedRecipeBuilder.shaped(RecipeCategory.TOOLS, FeruchemyItems.NICROSIL_METAL_NECKLACE).define('X', inputMaterial).pattern("XXX").pattern("X X").pattern(" X ").group("necklace").unlockedBy("has_material", has(inputMaterial)).save(consumer);
```
leafreynolds: because all these sections are basically the same as their original. It means you're not reusing code efficiently

src/api/java/leaf/cosmere/api/Manifestations.java
leafreynolds: Specifically, don't add hemalurgy as an ID here. It isn't a power type in and of itself. You also changed all the numbers of the other powers? which can potentially affect current saves, depending on how far we went with that. it's not good practice.

RequestingKidney:
Comments are acknowledged. I’m going to review and clean them up. Tbh, the way it was implemented was a real chore without a common interface to abstract granting manifestations across modules. I might pull it back and wait to properly demo that.

Gerbagel:
Starting a server crashes, here's the log: 

https://mclo.gs/xviOMOY

This could just be me, of course. 

More concrete, we usually allow inventory metalminds as well, while your nicrosil only seems to allow metalminds equipped in curio slots. It should be consistent with the rest of the mod, in my opinion. If it makes your UI work harder, then maybe we can use this in the interim, just have to ask Leaf lol

RequestingKidney:
1. OOPS I should’ve done better server testing. I’ll review that after work today. 
2. Id love to have a discussion on discord on whether or not metalminds in inventories can be used.

RequestingKidney:
@Gerbagel The fixes are in! Metalminds in the inventory now work, and server now runs. I forgot to put a "client only" decorator for the openMenu function.

Gerbagel:
Alright, checked so the server and inventory metalminds work. Looks good to me, even actually read the changes lol. I usually let Leaf do the merge themselves, so it might take a while, but you have my approval

src/feruchemy/java/leaf/cosmere/feruchemy/client/gui/NicrosilMenu.java
```diff
@@ -0,0 +1,1227 @@
...
+        }
+
+        fromSlot.setAttribute(null);
+        fromSlot.setStrength(0);
+
+        playerSpiritwebPowerButtons.clear();
+        sidedMenuButtons.clear();
+        availableAttributes.removeIf(att -> att.getAttribute() == attribute);
+        availableAttributes.add(spiritweb.getLiving().getAttribute(attribute));
+        spiritweb.getLiving().getAttribute(attribute).setBaseValue(strength);
```
leafreynolds: All menus are understandably client side only 😄 but trying to update the attribute of the entity is a server side only operation. 

If you're wanting the server to do something like this, you'll have to fire off a packet to the server, then have the server sync the changes back to the client. Server should auto sync attribute only changes already.

I recommend using the `runServer` run config in intellij (accepting the eula if needed and then running again), then having your `runClient` connect to localhost as the server

src/feruchemy/java/leaf/cosmere/feruchemy/client/gui/NicrosilMenu.java
```diff
@@ -0,0 +1,1227 @@
...
+            case SANDMASTERY:
+                // supported
+                break;
+            default:
+                // No icon for other types
+                return null;
+        }
+
+        if (type == ManifestationTypes.SANDMASTERY) {
+            int temp = 5;
```
leafreynolds: what does this code do?

leafreynolds (commented):
Feruchemical Nicrosil improvements review — the menu-driven approach, ISyncSpiritweb abstraction, and getPowers() API additions are solid. Main concern: the server-side tick logic was removed without a confirmed replacement path. See inline notes.

src/feruchemy/java/leaf/cosmere/feruchemy/common/manifestation/FeruchemyNicrosil.java
```diff
@@ -31,180 +19,28 @@ public FeruchemyNicrosil(Metals.MetalType metalType)
...
-			}
-		}
-		else
-		{
-			//this will clear call onModeChange and then clear nicrosil powers.
-			data.setMode(this, 0);
-		}
-
-		return false;
+		return 0;
```
leafreynolds, line 39: `modeMin` and `modeMax` both return `0`, disabling the Nicrosil mode system entirely — a player can never activate store/tap through the normal radial menu path. The old `tick()`, `checkStoreNicrosil`, and `checkTapNicrosil` logic has been deleted without a server-side replacement. The menu sends `StoreTapPowerMessage` packets — please confirm the server handler for those packets fully covers the removed tick behavior before merging, otherwise Nicrosil investiture is silently broken.

src/feruchemy/java/leaf/cosmere/feruchemy/client/gui/NicrosilMenu.java
leafreynolds: `setBaseValue` is a server-authoritative operation being called on the client side. This will cause visual desync with the actual server value and potentially corrupt local state. Client-side attribute previews should be read-only; the server should own all base value mutations and sync them back via packets.

src/feruchemy/java/leaf/cosmere/feruchemy/client/gui/NicrosilMenu.java
leafreynolds: Dead debug code — `int temp = 5` does nothing. Remove before merge.

src/feruchemy/java/leaf/cosmere/feruchemy/common/items/NicrosilBraceletMetalmindItem.java
```diff
@@ -0,0 +1,38 @@
...
+
+	@Override
+	public int getMaxCapacity()
+	{
+		return 2;
+	}
+
+	@Override
+	@OnlyIn(Dist.CLIENT)
+	public void appendHoverText(ItemStack stack, Level worldIn, List<Component> tooltip, TooltipFlag flagIn)
```
leafreynolds: This `appendHoverText` override just calls `super` — it adds nothing and should be removed. Same applies to `NicrosilNecklaceMetalmindItem`.

src/api/java/leaf/cosmere/api/manifestation/Manifestation.java
```diff
@@ -130,6 +131,15 @@ public Manifestation getManifestation()
 		return this;
 	}
 
+	public boolean hasMenu()
+	{
+		return false;
+	}
+
+	public void openMenu(Minecraft minecraft)
```
leafreynolds: `openMenu` takes `net.minecraft.client.Minecraft` — a client-only class — in a method defined on the `api` source set, which per project convention should remain dependency-free from client implementations. The empty default body means it will not crash server-side, but this is a layering violation. Consider moving this to a client-only interface or using `@OnlyIn(Dist.CLIENT)` with care.

leafreynolds: not sure how much we should care about this comment. it's not exactly wrong, but we already misuse the intention of the api domain

Gerbagel: I think it's probably best ignored. Maybe there's a better way to do this that complies with the intent of the api domain, but we're already in this deep lol

src/feruchemy/java/leaf/cosmere/feruchemy/client/gui/NicrosilMenu.java
Gerbagel: Perhaps it would be better to send a message to the server to handle the storing and tapping aspects entirely rather than just an attribute, but we fix that later so this can finally get merged. Unless you *want* to, of course, won't stop you :)

src/feruchemy/java/leaf/cosmere/feruchemy/common/manifestation/FeruchemyNicrosil.java
Gerbagel: Makes sense for f-nicrosil to be this way, since it's storing powers discretely. There's no need for any of these functions, I believe, since it's all handled via menu interactions now

src/feruchemy/java/leaf/cosmere/feruchemy/client/gui/NicrosilMenu.java
RequestingKidney: I do have the server handle this, and its via the `StoreTapPowerMessage`. Waiting for the client to sync to the server when moving powers between player and metalmind was an issue. I made sure the client had the intended state that was being sent to the server so players wouldn't be blocked. Each of those applyLocal functions are called after sending a message to the server.

src/feruchemy/java/leaf/cosmere/feruchemy/client/gui/NicrosilMenu.java
RequestingKidney: Damn. forgot about that. Removing it.

src/feruchemy/java/leaf/cosmere/feruchemy/common/manifestation/FeruchemyNicrosil.java
RequestingKidney: In the radial menu, there's now logic to open a menu rather than firing a packet to increase or decrease the mode. There's no mode for this manifestation now, so its appropriate to set the min and max to 0. 

There is no more tick behavior. Nicrosil store and tap is discrete. StoreTapPowerMessage handles the storing and tapping.

src/feruchemy/java/leaf/cosmere/feruchemy/client/gui/NicrosilMenu.java
RequestingKidney: duplicate

src/feruchemy/java/leaf/cosmere/feruchemy/common/items/NicrosilBraceletMetalmindItem.java
RequestingKidney: Easy fix, got it.

src/feruchemy/java/leaf/cosmere/feruchemy/client/gui/NicrosilMenu.java
RequestingKidney: See the other comment. The syncing of the server back to the client via Server still owns it, and will override the client if the client-side logic I implemented applied incorrectly. Sending packet to server and syncing to client via SpiritwebCapability's syncToClients method caused noticeable delay in ability to stop and tap powers in the menu.

src/api/java/leaf/cosmere/api/manifestation/Manifestation.java
RequestingKidney: Lmk what you want me to do with this one.

src/feruchemy/java/leaf/cosmere/feruchemy/client/gui/NicrosilMenu.java
RequestingKidney: Huh, seems to work just fine without it now. I'll keep testing but I removed this line. Thx.

src/feruchemy/java/leaf/cosmere/feruchemy/client/gui/NicrosilMenu.java
RequestingKidney: It's done, resolve when you're ready.

src/feruchemy/java/leaf/cosmere/feruchemy/client/gui/NicrosilMenu.java
RequestingKidney: Removed the line. Seems to be working fine. Can't remember why I had this issue before but I must've resolved it and this is legacy crap. Resolve when you're ready.

src/feruchemy/java/leaf/cosmere/feruchemy/common/items/NicrosilBraceletMetalmindItem.java
RequestingKidney: It's done, resolve when you're ready.

src/api/java/leaf/cosmere/api/manifestation/Manifestation.java
```diff
@@ -8,6 +8,7 @@
 import leaf.cosmere.api.Manifestations;
 import leaf.cosmere.api.providers.IManifestationProvider;
 import leaf.cosmere.api.spiritweb.ISpiritweb;
+import net.minecraft.client.Minecraft;
```
leafreynolds: **Critical — Dedicated Server Crash**

This import will cause `NoClassDefFoundError` at class-load time on a dedicated server. `net.minecraft.client.Minecraft` does not exist in the server jar, and `Manifestation` is a shared-API class loaded on both sides. The `@OnlyIn` on the override in `FeruchemyNicrosil` does not protect against this — the import in *this* class is enough to crash the server.

Remove `openMenu(Minecraft)` from the API. The `SpiritwebMenu` already has the `hasMenu()` branch and can open the correct screen from purely client-side code.

src/main/java/leaf/cosmere/common/charge/IHoldsPowers.java
```diff
@@ -0,0 +1,268 @@
...
+	{
+		if (itemStack.getItem() instanceof IHasMetalType metalType && metalType.getMetalType() == Metals.MetalType.NICROSIL)
+		{
+			UUID attunedPlayerID = getAttunedPlayer(itemStack);
+			UUID playerID = entity.getUUID();
+			boolean noAttunedPlayer = attunedPlayerID == null;
+
+			if (noAttunedPlayer)
+			{
+				ISpiritweb spiritweb = SpiritwebCapability.get(entity).resolve().get();
```
leafreynolds: `resolve().get()` without a prior `isPresent()` check will throw `NoSuchElementException` if the capability is absent (e.g. non-player entity, or cap not yet attached during early ticks). The companion method `getPlayerIsAttuned` guards with `isPresent()` first — this one should too:

```java
SpiritwebCapability.get(entity).ifPresent(spiritweb -> {
    CosmereEffect aluminumEffect = ...;
    if (spiritweb.hasEffect(aluminumEffect)) { ... }
});
```

src/main/java/leaf/cosmere/common/network/packets/StoreTapPowerMessage.java
```diff
@@ -0,0 +1,103 @@
...
+				}
+				// Tapping
+				else
+				{
+                    ItemStack itemStack = CosmereAttributeUtils.getPowerItem(sender, itemSlot, isCurio);
+                    if (itemStack.getItem() instanceof IHoldsPowers item)
+                    {
+                        if(item.getPlayerIsAttuned(itemStack, sender))
+                        {
+                            CosmereAttributeUtils.grantBaseAttribute(sender, (RangedAttribute) attribute, attributeStrength);
```
leafreynolds: Unconditional cast to `RangedAttribute` will throw `ClassCastException` on the server if a client sends a non-ranged attribute (or a malformed packet). Add a guard:

```java
if (attribute instanceof RangedAttribute ranged) {
    CosmereAttributeUtils.grantBaseAttribute(sender, ranged, attributeStrength);
}
```

leafreynolds (changes requested):
## Review: Client vs Server Side Issues

Three issues need fixing before merge, one of which is a guaranteed dedicated-server crash.

### Critical — Dedicated Server Crash

`Manifestation.java` (API) now imports `net.minecraft.client.Minecraft` and declares `openMenu(Minecraft)`. Because `Manifestation` lives in `src/api` and is loaded on both sides, the JVM will throw `NoClassDefFoundError` at class-load time on a dedicated server — no matter that the method body is empty and the override has `@OnlyIn`.

**Fix:** Remove `openMenu(Minecraft)` from the API entirely. `SpiritwebMenu.mouseClicked` already has the `hasMenu()` branch; it can call `NicrosilMenu` or any other screen directly there, keeping all client-class references out of the shared API.

### Moderate — Unsafe `.get()` Without `isPresent()` (Server-side)

`IHoldsPowers.trySetAttunedPlayer` calls `SpiritwebCapability.get(entity).resolve().get()` unconditionally. If the capability is absent (non-player entity, or capability not yet attached) this throws `NoSuchElementException` on the server. The companion method `getPlayerIsAttuned` does guard with `isPresent()` first — `trySetAttunedPlayer` should do the same.

### Moderate — Unchecked Cast to `RangedAttribute` (Server-side)

`StoreTapPowerMessage.handle` does an unconditional `(RangedAttribute) attribute` cast. The attribute arrives as a description-ID string from the client, so any malformed or unexpected packet would cause a `ClassCastException` on the server. Add an `instanceof` guard before casting.

src/main/java/leaf/cosmere/common/charge/IHoldsPowers.java
leafreynolds: It's wrong in this case, because we want to return true out of this function, which we can't do with ifPresent lambda

src/api/java/leaf/cosmere/api/manifestation/Manifestation.java
leafreynolds: sure why not.

PR #201, opened 2025-12-26
Fix Copper Block Decompression Issue

RequestingKidney:
- Changed copper ingot recipe to be under cosmere recipe resources and not minecraft's.


Gerbagel:
Feels like this should've broken either when Leaf removed cosmere:copper a long time ago, or when I did some stuff with copper that I can't recall the nature of, so weird no one noticed till now. 

Anyway, I see no issues.

RequestingKidney:
I agree. I haven’t loaded an older version to confirm if this occurs in older instances. It seems, though, that this code for generating the recipe has been unchanged for a while.

Gerbagel:
I stand corrected, I see issues lol

image

Gerbagel:
Oh right, I need to manually switch branches, woops

Gerbagel:
Yep, works, feels kinda wack to make copper a Cosmere recipe but whatever

Issue #162, opened 2025-08-03
[Surgebinding] Most surges have _some_ use


leafreynolds:

leafreynolds:
Resolved by #202.

Split out Transportation to #218.

All surges will eventually need a polish pass.

Issue #161, opened 2025-08-03
[Surgebinding] Ability to get viable stormlight capable gemstones in survival


leafreynolds:

leafreynolds:
Resolved by #202

PR #202, opened 2026-01-19
Surgebinding Implementation

CrazyBookEnthusiast:
Added Right Click on Block functionality for Division
Added application of Self for Abrasion (Need to make Effects for Sticking and Slicking)


CrazyBookEnthusiast:
Alpha state achieved for all Surges other than Cohesion, Tension and Transformation
Added Ideal Managers for some more Orders.

Need to Do:
-Ideal Progression via XP
-Teleport via Transportation
-Transformation, Cohesion, Tension
-Infusing Blocks for Abrasion and Adhesion
-Illusion for Illumination

Gerbagel:
For some reason, when swearing ideals, the player's spiritweb is null. Is it being passed correctly? 

Also, if you're going to try your hand at the Ideal Progression with XP thing that we've been discussing in the Discord, do it with another pull request. Makes things easier to sort through and handle is all.

CrazyBookEnthusiast:
#161 
RESOLVED
I just need Budding Sapphire but other than that, all the gems spawn as ores and you can stonecut them into Stormlight Viable Gemstones

CrazyBookEnthusiast:
#162 Every Surge has Some use
So the uses are following
Abrasion: Gives you Speed or Slowness, will do the whole friction thing
Adhesion: Punch a creature and Slowness 50, you can also punch them again to regain the stormlight
Cohesion: You can break blocks that would be stoneshapeable
Division: Wither on hit and also break blocks but they don't drop
Gravitation: haven't changed anything
Illumination: Gives you nightvision (This will not be staying in any form, I just didn't know how to do this)
Progression: haven't touched it
Tension: If you shift while active, you get resistance
Transformation: REALLY clunky essence transformation
Transportation: Peering into Shadesmar and set up ready for Selecting multiple entities around you

CrazyBookEnthusiast:
@leafreynolds This PR is ready for Review.

src/surgebinding/java/leaf/cosmere/surgebinding/common/eventHandlers/SurgebindingForgeEventsHandler.java
```diff
@@ -81,6 +83,16 @@ public static void onLivingAttackEvent(LivingAttackEvent event)
 		}
 
 		SurgeGravitation.onLivingAttackEvent(event);
+		SurgeDivision.onLivingAttackEvent(event);
+		SurgeAdhesion.onLivingAttackEvent(event);
+	}
+
+	@SubscribeEvent
+	public static void onShift(MovementInputUpdateEvent event){
```
leafreynolds: **[Critical — Server Crash]** `MovementInputUpdateEvent` is a client-only class (`net.minecraftforge.client.event`), but `SurgebindingForgeEventsHandler` has no `value = Dist.CLIENT` restriction on its `@Mod.EventBusSubscriber`. When a dedicated server loads this class, it throws `NoClassDefFoundError` because the event class does not exist on the server classpath.

Fix: add `value = net.minecraftforge.api.distmarker.Dist.CLIENT` to the annotation, or move `onShift` into a separate client-only handler class.

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/ideals/RadiantStateManager.java
```diff
@@ -220,7 +220,7 @@ private boolean onTrySuccessfulSwear(int idealToSwear, Roshar.RadiantOrder ideal
 		{
 			case WINDRUNNER ->
 			{
```
leafreynolds: **[Critical — Ideal Progression Broken]** This PR removes the `return WindrunnerIdealStateManager.validateIdeal(spiritweb, idealToSwear)` call and replaces all switch cases with empty blocks. `onTrySuccessfulSwear` now always falls through to `return false` for any ideal beyond the first — players can swear the first ideal but can never advance further regardless of order.

The new `*XPManager` classes added in this PR all have `validateIdeal` hardcoded to `return false` and are never called. At minimum, each case should delegate to `[Order]XPManager.validateIdeal(spiritweb, idealToSwear)` to keep the wiring in place.

leafreynolds: can probably safely ignore this one

src/surgebinding/java/leaf/cosmere/surgebinding/common/eventHandlers/SurgebindingForgeEventsHandler.java
leafreynolds: I think we do have a client side event handler that this could go to

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeDivision.java
```diff
@@ -14,5 +33,63 @@ public SurgeDivision(Roshar.Surges surge)
...
+					}
+				}
+				else
+				{
+					if (submodule.adjustStormlight(-15, true))
+					{
+						if (event.getLevel() instanceof ServerLevel serverLevel)
+						{
+							serverLevel.destroyBlock(blockPos, false);
+							serverLevel.addParticle(ParticleTypes.CAMPFIRE_COSY_SMOKE, blockPos.getX(), blockPos.getY(), blockPos.getZ(), 0.0D, 0.1D, 0.0D);
```
leafreynolds: `ServerLevel.addParticle()` is a no-op on the server — particles spawned this way are silently dropped. Use `serverLevel.sendParticles(ParticleTypes.CAMPFIRE_COSY_SMOKE, x, y, z, count, dx, dy, dz, speed)` instead to broadcast to nearby clients.

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeTransportation.java
```diff
@@ -13,7 +31,105 @@ public SurgeTransportation(Roshar.Surges surge)
...
+					entity.addEffect(EffectsHelper.getNewEffect(MobEffects.GLOWING,9,4));
+				}
+			}
+
+		});
+
+		return super.tick(data);
+	}
+
+	static List<Entity> targets = new LinkedList<>();
```
leafreynolds: **[Multiplayer Bug]** `targets` and `shiftDuration` are static class fields, meaning they are shared across all players on the server. Player A holding shift will overwrite Player B's charge state and target list.

Both fields need to be stored per-player — either in the `SurgebindingSpiritwebSubmodule` (preferred, since it already tracks per-player surge state) or in a `Map<UUID, ...>` keyed by player UUID.

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeTransportation.java
```diff
@@ -13,7 +31,105 @@ public SurgeTransportation(Roshar.Surges surge)
...
+		{
+			if (iSpiritweb.hasManifestation(SurgebindingManifestations.SURGEBINDING_POWERS.get(Roshar.Surges.TRANSPORTATION).get()) &&
+				SurgebindingManifestations.SURGEBINDING_POWERS.get(Roshar.Surges.TRANSPORTATION).getManifestation().isActive(iSpiritweb) &&
+				iSpiritweb.getMode(SurgebindingManifestations.SURGEBINDING_POWERS.get(Roshar.Surges.TRANSPORTATION).getManifestation())>=3)
+			{
+				SurgebindingSpiritwebSubmodule submodule = (SurgebindingSpiritwebSubmodule) iSpiritweb.getSubmodule(Manifestations.ManifestationTypes.SURGEBINDING);
+				if(event.getInput().shiftKeyDown){
+					if(!(submodule.getIdeal()>3)){
+						if(shiftDuration==10){
+							shiftDuration=10;
```
leafreynolds: Dead no-op code — `shiftDuration == 10` only runs when it's already 10, and then sets it to 10. If the intent was to cap `shiftDuration` at 10 for players below ideal 4, the condition should be `if (shiftDuration >= 10) { shiftDuration = 10; }` placed after `shiftDuration++` in `chargeUp`.

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeTransportation.java
```diff
@@ -13,7 +31,105 @@ public SurgeTransportation(Roshar.Surges surge)
...
+						}
+					}
+					chargeUp(event);
+				}
+				else{
+					if(shiftDuration>0){
+						shiftDuration=0;
+						System.out.println(targets);
```
leafreynolds: Debug `System.out.println` left in. Replace with `CosmereAPI.logger.debug(...)` or remove.

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeTransportation.java
```diff
@@ -13,7 +31,105 @@ public SurgeTransportation(Roshar.Surges surge)
...
+						if(!(submodule.getIdeal()>3)){
+							targets.clear();
+						}
+						targets.add(0, iSpiritweb.getLiving());
+						for(Entity target : targets){
+							if(submodule.adjustStormlight(60,true))
```
leafreynolds: Two issues here:

1. **Wrong sign** — every other surge passes a negative value to drain stormlight (e.g. `adjustStormlight(-15, true)`, `adjustStormlight(-40, true)`). `60` is positive, which by that convention would *grant* stormlight rather than consume it. Should this be `-60`?

2. **Client-side stormlight drain** — `onShift` is called from `MovementInputUpdateEvent`, which fires on the client. `adjustStormlight` here only modifies the client-side spiritweb; the server never sees the drain. Any actual gameplay effect should be triggered via a server-side packet.

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeTransportation.java
```diff
@@ -13,7 +31,105 @@ public SurgeTransportation(Roshar.Surges surge)
...
+							{
+								//Teleport Effect
```
leafreynolds: Teleportation is not implemented — only a comment placeholder. Stormlight is drained (or gained, see above) for each target but nothing actually happens. This block should either be implemented or guarded with a TODO exception/log so it's clear the feature is incomplete.

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeTransformation.java
```diff
@@ -13,7 +22,65 @@ public SurgeTransformation(Roshar.Surges surge)
...
+							event.getLevel().setBlock(blockPos, Blocks.OAK_LEAVES.defaultBlockState(), 1);
+							break;
+						case 6:
+							event.getLevel().setBlock(blockPos, Blocks.WATER.defaultBlockState(), 1);
+							break;
+						case 7:
+							event.getLevel().setBlock(blockPos, Blocks.ICE.defaultBlockState(), 1);
+							break;
+						case 8:
+							submodule.adjustStormlight(-100,true);
```
leafreynolds: **Double stormlight drain** — the outer `adjustStormlight(-50, true)` check at line 42 already deducted 50 stormlight before we get here. This additional `-100` makes the total cost for case 8 be 150, with no indication to the player. Consider using a per-mode cost variable instead of a hardcoded inner drain:
```java
int cost = surge.getMode(iSpiritweb) == 8 ? -150 : -50;
if (submodule.adjustStormlight(cost, true)) { ... }
```

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeTransformation.java
```diff
@@ -13,7 +22,65 @@ public SurgeTransformation(Roshar.Surges surge)
...
+				if(submodule.adjustStormlight(-50,true))
+				{
+					switch (surge.getMode(iSpiritweb))
+					{
+						case 1:
+							event.getLevel().setBlock(blockPos, Blocks.AIR.defaultBlockState(), 1);
+							break;
+						case 2:
+							event.getLevel().setBlock(blockPos, Blocks.AIR.defaultBlockState(), 1);
+							event.getLevel().addParticle(ParticleTypes.LARGE_SMOKE,blockPos.getX(),blockPos.getY(),blockPos.getZ(),0,0,0);
```
leafreynolds: `event.getLevel().addParticle()` is called from a `RightClickBlock` event which fires on both client and server. On the server, `addParticle()` is a no-op. Use `ServerLevel.sendParticles()` for server-side particle spawning (same fix needed in `SurgeDivision.java`).

src/surgebinding/java/leaf/cosmere/surgebinding/common/capabilities/ideals/order/WindrunnerXPManager.java
```diff
@@ -0,0 +1,20 @@
+package leaf.cosmere.surgebinding.common.capabilities.ideals.order;
+
+import leaf.cosmere.common.cap.entity.SpiritwebCapability;
+
+public class WindrunnerXPManager
+{
+	public int wrXP = 0;
+	public static boolean validateIdeal(SpiritwebCapability spiritweb, int ideal){
+		//Implement XP thresholds
```
leafreynolds: All nine `*XPManager` classes (`WindrunnerXPManager`, `SkybreakerXPManager`, etc.) are identical empty stubs — `validateIdeal` always returns `false`, and `xpUp`/`xpDown` do nothing. None of them are called from `RadiantStateManager`.

Combined with the empty switch in `onTrySuccessfulSwear`, this means ideal progression is completely non-functional. If these are placeholders for a follow-up PR, the switch cases in `RadiantStateManager` should at minimum call `validateIdeal` so the wiring compiles and only the threshold logic is missing.

leafreynolds: can probably ignore this one

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeTransportation.java
```diff
@@ -13,7 +31,105 @@ public SurgeTransportation(Roshar.Surges surge)
...
+				AABB areaEffect = new AABB(new Vec3(living.getX()-3, living.getY()-3, living.getZ()-3), new Vec3(living.getX()+3, living.getY()+3, living.getZ()+3));
+				List<Entity> entitiesNear = living.level().getEntities(living,areaEffect);
+				List<LivingEntity> entityList= new LinkedList<>();
+				for(Entity entity : entitiesNear){
+					if(entity instanceof LivingEntity){
+						entityList.add((LivingEntity)entity);
+					}
+				}
+				for(LivingEntity entity : entityList){
+					entity.addEffect(EffectsHelper.getNewEffect(MobEffects.GLOWING,9,4));
```
leafreynolds: `GLOWING` is re-applied to every nearby entity on every tick with no cooldown or active-check. `addEffect` called each tick will keep resetting the effect duration, which is fine, but the AABB query + list allocation + effect application runs every tick unconditionally while Transportation is active. At minimum check `!entity.hasEffect(MobEffects.GLOWING)` before calling `addEffect`, or use a tick counter to throttle to once per second.

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeAbrasion.java
```diff
@@ -13,7 +20,35 @@ public SurgeAbrasion(Roshar.Surges surge)
...
+		{
+			return false;
+		}
+		int mode = getMode(data);
+		LivingEntity livingEntity = data.getLiving();
+		SurgebindingSpiritwebSubmodule submodule = (SurgebindingSpiritwebSubmodule) data.getSubmodule(Manifestations.ManifestationTypes.SURGEBINDING);
+		if(data.hasManifestation(SurgebindingManifestations.SURGEBINDING_POWERS.get(Roshar.Surges.ABRASION).getManifestation()) &&
+			SurgebindingManifestations.SURGEBINDING_POWERS.get(Roshar.Surges.ABRASION).getManifestation().isActive(data))
+		{
+			if (mode>0 && submodule.adjustStormlight(-mode,true))
```
leafreynolds: `mode` is used simultaneously as both the stormlight drain rate and the effect amplifier. These two concerns scale together by coincidence rather than design:
- `adjustStormlight(-mode, true)` — drains `mode` stormlight per tick  
- `addEffect(MobEffects.MOVEMENT_SPEED, mode, 2)` — sets amplifier to `mode`

A player at a high ideal will be drained far more stormlight per tick than intended, and a player at low mode gets almost no speed boost. Stormlight cost per tick and effect amplifier should be independent values.

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeCohesion.java
```diff
@@ -17,4 +33,51 @@ public SurgeCohesion(Roshar.Surges surge)
...
+	}
+
+	public static void onBlockInteract(PlayerInteractEvent.RightClickBlock event){
+		final BlockPos blockPos = event.getHitVec().getBlockPos();
+		Block block = event.getLevel().getBlockState(blockPos).getBlock();
+		SpiritwebCapability.get(event.getEntity()).ifPresent(iSpiritweb ->
+		{
+			SurgebindingSpiritwebSubmodule submodule = (SurgebindingSpiritwebSubmodule) iSpiritweb.getSubmodule(Manifestations.ManifestationTypes.SURGEBINDING);
+			SurgebindingManifestation surge = (SurgebindingManifestation) SurgebindingManifestations.SURGEBINDING_POWERS.get(Roshar.Surges.COHESION).getManifestation();
+			if(iSpiritweb.hasManifestation(surge) && isValidStoneBlock(block))
```
leafreynolds: There's a lot of random files that seem to be missing spaces in places.

I think if you tell it to format the code, it should automatically be using our formatting style. Not a huge issue, but it's nice for code styles to be consistent.

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeDivision.java
```diff
@@ -14,5 +33,63 @@ public SurgeDivision(Roshar.Surges surge)
 	}
 
 	//power over destruction and decay
+	public static void onBlockInteract(PlayerInteractEvent.RightClickBlock event)
+	{
+		final BlockPos blockPos = event.getHitVec().getBlockPos();
+
+		SpiritwebCapability.get(event.getEntity()).ifPresent(iSpiritweb ->
+		{
+			SurgebindingManifestation surge = (SurgebindingManifestation) SurgebindingManifestations.SURGEBINDING_POWERS.get(Roshar.Surges.DIVISION).getManifestation();
```
leafreynolds: Unchecked cast — if `getManifestation()` ever returns a non-`SurgebindingManifestation` object (e.g. due to a registry error or future refactor), this silently throws `ClassCastException`. The same pattern appears in `SurgeTransformation.onBlockInteract` and `SurgeCohesion.onBlockInteract`. Consider adding an `instanceof` guard:
```java
if (SurgebindingManifestations.SURGEBINDING_POWERS.get(Roshar.Surges.DIVISION).getManifestation() instanceof SurgebindingManifestation surge) { ... }
```

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeTransformation.java
```diff
@@ -13,7 +22,65 @@ public SurgeTransformation(Roshar.Surges surge)
 		super(surge);
 	}
 
-
 	//soulcasting, changing one thing into another
 
+
+	@Override
+	public int modeMax(ISpiritweb data)
```
leafreynolds, line 30: `modeMax` is hardcoded to `10`, bypassing the ideal-based gating in `SurgebindingManifestation`. The base class ties `modeMax` to `spiritweb.getIdeal()`, so players are only supposed to access higher modes as they swear more ideals. Returning a fixed 10 here gives all players access to every transformation mode from the start, regardless of their current ideal.

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeTransportation.java
leafreynolds: a throttle isn't a bad suggestion. We probably don't want to be doing huge scanning checks every tick, bad for performance

src/surgebinding/java/leaf/cosmere/surgebinding/common/manifestation/SurgeAbrasion.java
leafreynolds: extra notes since the ai comment probably doesn't know. An amplifier of zero is just a normal effect.

leafreynolds:
This was the full output of the ai code review in case it's helpful. Not everything in it specifically matters, but may be helpful as a tool for finding things you potentially missed.

```
  Critical Bugs

  1. Dedicated server crash — MovementInputUpdateEvent in a non-@Dist.CLIENT subscriber

  SurgebindingForgeEventsHandler.java line 91 subscribes to MovementInputUpdateEvent, which is a client-only class (net.minecraftforge.client.event). The class has no value = Dist.CLIENT restriction on @Mod.EventBusSubscriber. When the dedicated server loads this      
  class, it will throw NoClassDefFoundError because the client event class doesn't exist on the server classpath. This crashes every dedicated server.

  Fix: Add value = net.minecraftforge.api.distmarker.Dist.CLIENT to the annotation, or extract onShift into a separate client-side event handler class.

  2. onTrySuccessfulSwear always returns false for ideals > 1 (RadiantStateManager.java lines 219–256)

  Every case in the switch is an empty block. The method falls through to return false for any ideal beyond the first. Players can never advance past the 1st ideal regardless of order.

  3. All XPManager classes are empty stubs and not wired up

  WindrunnerXPManager, SkybreakerXPManager, and all other *XPManager classes have validateIdeal hardcoded to return false and empty xpUp/xpDown methods. They are never called from RadiantStateManager. If the intent is to gate these behind a separate PR, the
  onTrySuccessfulSwear switch should at least call validateIdeal so the wiring is in place.

  4. serverLevel.addParticle() doesn't work server-side (SurgeDivision.java line 67)

  ServerLevel.addParticle() is a no-op. Use serverLevel.sendParticles(...) instead. Same issue exists in SurgeTransformation.java line 51 — event.getLevel().addParticle() is called from a RightClickBlock handler that fires on both sides; on the server it does nothing. 

  ---
  Logic Bugs

  5. SurgeTransportation — static fields shared across all players (lines 64–65)

  static List<Entity> targets = new LinkedList<>();
  static int shiftDuration = 0;

  These are class-level statics. In multiplayer, player A holding shift will overwrite player B's shiftDuration and targets. Both fields must be stored per-player (e.g., in the spiritweb submodule or a Map<UUID, ...>).

  6. SurgeTransportation.onShift — stormlight consumed client-side (line 93)

  onShift is called from MovementInputUpdateEvent, which fires on the client. submodule.adjustStormlight(60, true) therefore modifies the client-side spiritweb only — the server never sees this drain. Also note the sign: 60 is positive, which by the mod's convention   
  (other surges use negative values like -15, -40) would add stormlight rather than consume it.

  7. SurgeTransportation — teleportation is not implemented (line 95)

  The "charge up and release to teleport" mechanic has //Teleport Effect as a comment with no actual code. The stormlight is still drained (or gained, see above) without doing anything.

  8. Dead / no-op code in SurgeTransportation.onShift (lines 77–80)

  if (!(submodule.getIdeal() > 3)) {
      if (shiftDuration == 10) {
          shiftDuration = 10;  // assigns variable to itself
      }
  }

  This is a no-op. The intent appears to be capping shiftDuration but the condition is never effective.

  9. SurgeAbrasion — mode used as both stormlight cost and effect amplifier

  mode drives both the cost of the surge and the amplifier of MOVEMENT_SLOWDOWN. These two concerns scale together, which isn't intentional design — a higher ideal shouldn't necessarily slow you more. Cost and amplifier should be independently configurable.

  10. SurgeTransformation case 8 — double stormlight drain (line 69)

  The outer check already deducted 50 stormlight (adjustStormlight(-50, true)). Inside case 8, another adjustStormlight(-100, true) is called. The total cost is 150 with no indication to the player. The 100 should probably replace the 50, using a separate cost variable
   per-mode.

  ---
  Code Quality Issues

  11. System.out.println left in (SurgeTransportation.java line 87)

  System.out.println(targets);

  Use CosmereAPI.logger or remove.

  12. Unchecked casts to SurgebindingManifestation (multiple files)

  SurgebindingManifestation surge = (SurgebindingManifestation) SurgebindingManifestations.SURGEBINDING_POWERS.get(Roshar.Surges.X).getManifestation();

  This pattern appears in SurgeTransformation, SurgeDivision, SurgeCohesion. If the registry ever returns a non-SurgebindingManifestation object, this silently ClassCastExceptions. Add an instanceof guard.

  13. SurgeTransportation.tick applies GLOWING to all nearby entities every tick with no cooldown

  Any entity within 3 blocks gets GLOWING applied every tick, which is both expensive and will constantly re-apply the effect. This should only be applied when the effect isn't already active, or use a tick counter.

  14. SurgeTransformation.modeMax returns 10 unconditionally

  @Override
  public int modeMax(ISpiritweb data) { return 10; }

  The base class ties modeMax to ideal level. Hardcoding 10 gives all players access to all transformation modes regardless of their sworn ideals.

```

CrazyBookEnthusiast:
I just want to say that I'm so happy that none of the world gen stuff had a problem because omg, I was not ready to attempt to fix that. Also I've started with fixes and will put it all in a commit in like 12 hrs

CrazyBookEnthusiast:
To be done:
Transforportation MovementInputEvent
Transportation static fields shared across players
Transformation logic fixes

PR #205, opened 2026-01-30
Moves power save state toggles to Server

SoaringEaqle:

New SavePowerStateMessage.java class
New TogglePowerStateMessage.java class
New SpiritwebCapability.PowerSaveState inner class ClientForgeEvents#onKey() now sends a single packet to the server for toggling Power Save States instead of potentially multiple


src/main/java/leaf/cosmere/common/cap/entity/SpiritwebCapability.java
```diff
@@ -205,9 +242,31 @@ public void deserializeNBT(CompoundTag compoundTag)
 		{
 			spiritwebSubmodule.deserialize(this);
 		}
-		if(nbt.contains("PowerSaveStates"))
+		if (nbt.contains("PowerSaveStates"))
 		{
-			PowerSaveState.deserialize((CompoundTag) nbt.get("PowerSaveStates"));
+			for(PowerSaveState state : powerSaveStorage)
+			{
+				CompoundTag data = (CompoundTag) nbt.get(Integer.toString(state.num));
```
leafreynolds: You're trying to read from 'nbt' variable directly, instead of reading in "PowerSaveStates" to a variable, then accessing the state from that new variable

src/main/java/leaf/cosmere/common/network/packets/TogglePowerStateMessage.java
```diff
@@ -0,0 +1,44 @@
...
+
+
+	@Override
+	public void handle(NetworkEvent.Context context)
+	{
+		ServerPlayer sender = context.getSender();
+		MinecraftServer server = sender.getServer();
+		server.submitAsync(() -> SpiritwebCapability.get(sender).ifPresent((data) ->
+		{
+			data.activatePowerState(powerState);
```
leafreynolds, line 28: This is being handled on the server, but when it calls `activatePowerState` ... it then calls `Cosmere.packetHandler().sendToServer` inside it. 

so it's the server trying to tell the server in a big ol loop?

src/main/java/leaf/cosmere/common/cap/entity/SpiritwebCapability.java
```diff
@@ -1032,4 +1101,116 @@ public void syncToClients(@Nullable ServerPlayer serverPlayerEntity)
 			Cosmere.packetHandler().sendTo(new SyncPlayerSpiritwebMessage(this.livingEntity.getId(), nbt), serverPlayerEntity);
 		}
 	}
+
+	public void saveNewState(int num)
+	{
+		powerSaveStorage[num].addManifestations(this);
```
leafreynolds, line 1080: we should probably handle index out of bounds errors here. Even if a client might never actually send an integer out of bounds, this is a public code base and people could be malicious, make their own jar file with a change that purposefully submits a silly number, and make the server have a bad time.

src/main/java/leaf/cosmere/common/cap/entity/SpiritwebCapability.java
```diff
@@ -1032,4 +1101,116 @@ public void syncToClients(@Nullable ServerPlayer serverPlayerEntity)
...
+	}
+
+	public void activatePowerState(int num)
+	{
+		powerSaveStorage[num].activate(this);
```
leafreynolds, line 1090: same issue here re index out of bounds.

src/main/java/leaf/cosmere/common/cap/entity/SpiritwebCapability.java
```diff
@@ -1032,4 +1101,116 @@ public void syncToClients(@Nullable ServerPlayer serverPlayerEntity)
...
+			{
+				int modifier = -(manifestation.getMode(spiritweb));
+				if (toActivate)
+				{
+					modifier += manifestations.get(manifestation);
+				}
+				Cosmere.packetHandler().sendToServer(new ChangeManifestationModeMessage(manifestation, modifier));
+
+			}
+			if (CosmereConfigs.CLIENT_CONFIG.disableActivatorChatMessage.get())
```
leafreynolds: if this activate function is being called from the server, then a client config isn't relevant.

src/main/java/leaf/cosmere/common/network/packets/SavePowerStateMessage.java
```diff
@@ -0,0 +1,42 @@
...
+import leaf.cosmere.common.cap.entity.SpiritwebCapability;
+import leaf.cosmere.common.network.ICosmerePacket;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraftforge.network.NetworkEvent;
+
+public class SavePowerStateMessage implements ICosmerePacket
+{
+	int powerState;
```
leafreynolds: should add private final to this

src/main/java/leaf/cosmere/common/network/packets/TogglePowerStateMessage.java
```diff
@@ -0,0 +1,44 @@
...
+import leaf.cosmere.common.cap.entity.SpiritwebCapability;
+import leaf.cosmere.common.network.ICosmerePacket;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraftforge.network.NetworkEvent;
+
+public class TogglePowerStateMessage implements ICosmerePacket
+{
+	int powerState;
```
leafreynolds: should add private final to this

src/main/java/leaf/cosmere/common/cap/entity/SpiritwebCapability.java
```diff
@@ -61,53 +62,74 @@
     https://coppermind.net/wiki/Ars_Arcanum#The_Alloy_of_Law
  */
 
+/**
+ * Capability-backed implementation of {@link ISpiritweb} for a living entity.
+ * Tracks manifestation ownership/modes, active Cosmere effects, and submodule state,
+ * and handles sync/serialization plus client rendering helpers.
+ */
```
leafreynolds: Kinda neat, but lets avoid adding AI written comments in the future. They're pretty low-value for they actually end up generating.

leafreynolds (changes requested):

leafreynolds:
Just a few lil changes needed :) good work.

Have you tried actually testing it in a server environment?

you can use `runServer` run config to start a server, (you may need to accept the eula text file it generates, then run again. Should only complain once.)

Afterwards, you can boot your client via `runClient`, and just connect to `localhost` as the server.

src/main/java/leaf/cosmere/common/cap/entity/SpiritwebCapability.java
SoaringEaqle: Could you explain what you're talking about here? I don't really understand.

leafreynolds: in serialize, you're saving each state under the "PowerSaveStates", on the nbt.
in deserialize, you are trying to read each state from nbt directly, instead of reading it from "PowerSaveStates" that's contained in nbt.

Nbt
-> powersavestates
--->0
--->1
--->2
--->3
--->4

etc

so you save stuff, but never read it back. So everything would get lost on game or server reboot

leafreynolds: How you're saving it
Image

src/main/java/leaf/cosmere/common/cap/entity/SpiritwebCapability.java
SoaringEaqle: Should I just make it return early if this occurs? Or should I do something specific (Throw a warning, log something, etc?)

leafreynolds: yep! I'd just ignore any non-valid indexes. you could also return an error, to the client, but I'm not sure that's worth it.

SoaringEaqle:
For whatever reason, testing the latest batch of changes in a server environment doesn't work for me. It keeps complaining about an invalid session. Don't know why. My running theory is it has something to do with me using a Mac, though I have little way of proving that.

leafreynolds:
> For whatever reason, testing the latest batch of changes in a server environment doesn't work for me. It keeps complaining about an invalid session. Don't know why. My running theory is it has something to do with me using a Mac, though I have little way of proving that.

do you have logs?

SoaringEaqle:
I think this is it. [https://mclo.gs/PmmppWR](https://mclo.gs/PmmppWR) Basically all the others are obfuscated, so I don't think they'll be much use.

SoaringEaqle:
Finished with this. It should be ready to go.

PR #214, opened 2026-03-27
[Feruchemy] Refactor metalmind selection and charge logic

### LINKED ISSUE #99: Refactor Metalmind Selection Code
The code that chooses what Metalmind to tap/store is one of the oldest pieces of code in the project, and likely needs a refactor to improve performance and be more intelligent about what order the metalminds will show in.

- [ ] Improve performance
- [ ] Improve Metalmind selection order (Hotbar, Curios, armor, inventory)

While we are in the area, we should:
- [ ] Fix bug that leaves only a few points left in a metalmind
- [ ] Allow grabbing points from multiple metalminds at once as part of one request (which may fix the above)

leafreynolds:
Closes #99

## Summary
- **Fix stranded charge bug**: multi-draw now spans multiple metalminds to satisfy a full cost; never partially applies (all-or-nothing)
- **Fix selection order**: hotbar → curios → armor → inventory (previously hotbar → inventory → armor → curios)
- **Two-pass design**: pass 1 is a read-only affordability check that collects accessible metalminds; pass 2 applies charge and attunement side-effects only if the full cost can be met
- **Dead code removal**: removed `adjustChargeExact` overloads with no external callers

## Test plan
- [ ] Tap a metal with charge spread across multiple metalminds — verify full cost is drawn and no charge is stranded
- [ ] Verify a tap that cannot be fully satisfied does not partially drain any metalmind
- [ ] Verify hotbar metalminds are preferred over inventory metalminds
- [ ] Verify curios slots are checked before armor and main inventory
- [ ] Verify attuned/unkeyed metalmind access rules still work correctly (own metalmind, unkeyed with/without identity)
- [ ] Verify storing identity (aluminum feruchemy) still allows storing in unkeyed metalminds
- [ ] Verify tin zoom render check (simulate=true) does not consume charge

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/feruchemy/java/leaf/cosmere/feruchemy/common/manifestation/FeruchemyManifestation.java
+++ b/src/feruchemy/java/leaf/cosmere/feruchemy/common/manifestation/FeruchemyManifestation.java
@@ -99,9 +99,10 @@ public class FeruchemyManifestation extends Manifestation implements IHasMetalTy
 	public boolean canAfford(ISpiritweb data, boolean simulate)
 	{
 		int adjustAmount = getCost(data);
-		final ItemStack metalmind = MetalmindChargeHelper.adjustMetalmindChargeExact(data, metalType, adjustAmount, !simulate, true);
+		boolean success = MetalmindChargeHelper.adjustMetalmindCharges(
+				data, metalType, adjustAmount, !simulate, true);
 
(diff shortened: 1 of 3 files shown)


leafreynolds:
@claude

PR #215, opened 2026-04-10
feature/1.20.x/migrate_spiritweb_to_tag

leafreynolds:

Spiritweb capable mobs are now assigned attributes and spiritweb based on the cosmere:has_spiritweb tag.

It means we can bypass the stupid java type erasure limitation, so other mods can technically provide support for spiritwebs if they so choose.


Gerbagel (commented):
Seems fine to me

src/main/java/leaf/cosmere/common/eventHandlers/EntityEventHandler.java
```diff
@@ -135,14 +132,16 @@ else if (eventEntity instanceof Warden warden)
 
 	public static boolean canStartWithPowers(Entity entity)
 	{
+		return entity.getType().is(CosmereTags.EntityTypes.HAS_SPIRITWEB);
```
leafreynolds: I'll need to fix this, I realised. We don't want cats and llamas starting with powers again

leafreynolds:
Pure datapack only is incapable of providing people the ability to add spiritwebs to whatever mobs they want, due to  when attributes are registered, versus when tags are actually loaded. 

Sorry folks.

PR #222, opened 2026-05-16
Better Combat Compat with Hemalurgy

RequestingKidney:

Better Combat Compat with Hemalurgy plus room for more later

## Testing checklist:

- [x] yes I tested this in a `runServer` environment, and therefore won't make leaf cry
- [x] yes, I actually connected to that server and verified it worked and don't need to have my kneecaps removed
- [no] yes, I also tested the built jars manually outside of the development environment, because problems sometimes don't show up inside the dev environment.


leafreynolds:
There's been a preference to not use the claude 1.21.1 port, so I will get you to switch the target branch to merge changes  over to `feature/1.21.1/human_port_1.21.1` :) sorry

PR #223, opened 2026-05-30
Feruchemy 1.21.1

RequestingKidney:

Work in progress

## Testing checklist:

- [ ] yes I tested this in a `runServer` environment, and therefore won't make leaf cry
- [ ] yes, I actually connected to that server and verified it worked and don't need to have my kneecaps removed
- [ ] yes, I also tested the built jars manually outside of the development environment, because problems sometimes don't show up inside the dev environment.


leafreynolds:
ty, copied what I needed 😄

PR #224, opened 2026-05-31
[Feruchemy] BugFix: Providing a Position in canSleep in FeruchemyBronze. 

EdgarVerdi:
- bug fix: provided a position in onPlayerSleepInBed of canSleep in FeruchemyBronze. This should prevent other mods from crashing when trying to access a non-null event.getPos() when using onTrySleep(PlayerSleepInBedEvent event)

## Testing checklist:

- [ ] yes I tested this in a `runServer` environment, and therefore won't make leaf cry
- [ ] yes, I actually connected to that server and verified it worked and don't need to have my kneecaps removed
- [x] yes, I also tested the built jars manually outside of the development environment, because problems sometimes don't show up inside the dev environment.


EdgarVerdi:
When testing with the runServer there seems to be an issue. However, it seems to also exist prior to my request. The error is related to a LocalPlayer class existing in a DEDICATED_SERVER. Hopefully I'm not missing something myself when checking this, however, this change is very small and likely didn't cause that.

